PR #13, opened 2022-04-02
support main files

Milo123459:
When using NPM/Yarn, in the package.json you can have the main entry. This now checks for that field, checks if the file exists and uses that as a start command if all else fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,4 +65,4 @@ jobs:
       - name: Run cargo clippy
         uses: actions-rs/cargo@v1
         with:
-          command: clippy
+          command: clippy --all-targets --all-features
--- a/src/providers/npm.rs
+++ b/src/providers/npm.rs
@@ -55,4 +55,5 @@ impl Provider for NpmProvider {
 pub struct PackageJson {
     pub name: String,
     pub scripts: Option<HashMap<String, String>>,
+    pub main: Option<String>,
 }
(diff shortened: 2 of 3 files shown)


coffee-cup (commented):
This is great, thank you! Can you add an integration test like the one here https://github.com/railwayapp/nixpacks/blob/main/tests/integration_tests.rs#L24.

I also need to figure out a clean way of sharing behaviour between providers but this is good for now.

Milo123459:
Cool, found a little caveat in `has_file` - it doesn't work for paths like `src/*`, so I'm going to work on adding that in this PR.

src/nixpacks/app.rs
```diff
@@ -23,13 +23,7 @@ impl App {
...
 
     pub fn includes_file(&self, name: &str) -> bool {
-        for path in &self.paths {
-            if path.file_name().unwrap() == name {
-                return true;
-            }
-        }
-
-        false
+        fs::read_to_string(self.source.join(name)).is_ok()
```
coffee-cup, line 26: Instead of reading the entire file into memory, I think it would be better to use something like [`fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) to simple check if the file exists on disk.

image

Something like

```rust
    pub fn includes_file(&self, name: &str) -> bool {
        let path = self.source.join(name);
        path.canonicalize().is_ok()
    }
```

coffee-cup (approved):
This is great, thanks!
Can you please merge in `main` just to get the tests to run. Then I will merge this

Milo123459:
Alright, I merged in. You need to approve the CI to run I think.

PR #15, opened 2022-04-04
Add Python provider

aleksrutins:
This PR adds a Python provider for Nixpacks.

I think it works. It was acting a little bit strangely - the Flask server is starting, but not responding correctly - but that's probably just me not knowing Flask and not passing the right flags to Docker.


Milo123459:
Make sure to add an integration test.

src/providers/python.rs
```diff
@@ -0,0 +1,37 @@
...
+use super::Provider;
+
+pub struct PythonProvider {}
+impl Provider for PythonProvider {
+    fn name(&self) -> &str {
+        "python"
+    }
+
+    fn detect(&self, app: &crate::nixpacks::app::App) -> anyhow::Result<bool> {
+        Ok(app.includes_file("main.py"))
```
Milo123459, line 14: Whilst most apps have this as their entry point, not all apps do. It should look for any python file.

src/providers/python.rs
```diff
@@ -0,0 +1,37 @@
...
+        }
+        Ok(None)
+    }
+
+    fn suggested_build_cmd(&self, _app: &App) -> Result<Option<String>> {
+        Ok(None)
+    }
+
+    fn suggested_start_command(&self, _app: &App) -> Result<Option<String>> {
+        Ok(Some("python main.py".to_string()))
```
Milo123459, line 35: This start command could also be wrong because main.py isn't always the entry point. It will need some more logic for detecting it.

src/providers/python.rs
FarazPatankar: Not well-versed with Python but might make more sense to check for a `requirements.txt` file?

Milo123459: `requirements.txt` doesn't have a main-file entry. That's why I suggested checking for all python files and seeing if there is one that makes sense to run. This could be heavily inspired by the Heroku one: https://github.com/heroku/heroku-buildpack-python

src/providers/python.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    fn pkgs(&self, _app: &crate::nixpacks::app::App) -> Vec<super::Pkg> {
+        vec![
+            Pkg::new("pkgs.python38")
+        ]
+    }
+
+    fn install_cmd(&self, app: &App) -> Result<Option<String>> {
+        if app.includes_file("requirements.txt") {
+            return Ok(Some("python -m ensurepip && python -m pip install -r requirements.txt".to_string()))
```
Milo123459, line 25: There is another way of specifying dependencies, using a `Pipfile`.

src/providers/python.rs
```diff
@@ -0,0 +1,37 @@
...
+        "python"
+    }
+
+    fn detect(&self, app: &crate::nixpacks::app::App) -> anyhow::Result<bool> {
+        Ok(app.includes_file("main.py"))
+    }
+
+    fn pkgs(&self, _app: &crate::nixpacks::app::App) -> Vec<super::Pkg> {
+        vec![
+            Pkg::new("pkgs.python38")
```
Milo123459, line 19: This could check if the `runtime.txt` file exists, which contains the Python version to use. Example: https://github.com/heroku/heroku-buildpack-python/blob/main/spec/fixtures/pipenv_and_runtime_txt/runtime.txt

src/providers/python.rs
coffee-cup: +1 to using Heroku buildpack as reference for detection. It looks like it checks for the exist of the 3 files and if any exist considers it a Python project.

image

src/providers/python.rs
Milo123459: Yeah, specific versioning of packages is going to be something quite hard to do.

coffee-cup (commented):
Thanks for the PR!

This project is still early so Python support does not have to be perfect right from the start, but there are a few things that I think should be added in this PR
- Check if python by looking for main.py or requirements.txt
- Check if `main.py` exists before setting it as the start command
- Integration tests and a Python example ([example](https://github.com/railwayapp/nixpacks/blob/main/tests/integration_tests.rs#L14))

Pipfile support would be great, but can easily come in a further PR by someone.

src/providers/python.rs
Milo123459: cc @coffee-cup - what do you think about this? Do you know how the Heroku buildpack knows which file to "run"?

aleksrutins:
I'm thinking that I should also add `pyproject.toml`/`setup.py`/`setup.toml` support, as that seems to be the recommended way to set up Python projects at this point.

aleksrutins:
This should now be ready.

aleksrutins:
Note that setuptools support currently only extends to projects configured entirely in `pyproject.toml`.

src/providers/python.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use super::Provider;
+
+pub struct PythonProvider {}
+impl Provider for PythonProvider {
+    fn name(&self) -> &str {
+        "python"
+    }
+
+    fn detect(&self, app: &crate::nixpacks::app::App, _env: &Environment) -> anyhow::Result<bool> {
```
coffee-cup, line 17:
```suggestion
    fn detect(&self, app: &crate::nixpacks::app::App, _env: &Environment) -> Result<bool> {
```

src/python/pyproject.rs
```diff
@@ -0,0 +1,69 @@
+use anyhow::{Context, Result};
+
+use crate::{chain, nixpacks::app::App};
+
+pub struct ProjectMeta {
```
coffee-cup, line 5: I think a lot of the chaining below can be cleaned up by deriving the Deserialize traits on this struct. Similar to the rust Cargo.toml format https://github.com/railwayapp/nixpacks/blob/main/src/providers/rust.rs#L13.

src/python/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod pyproject;
```
coffee-cup, line 1: Can this be moved into `providers/python`. Ideally there are no language specific modules outside of the providers directory. The core nixpacks logic and directory structure should be as language agnostic as possible.

src/python/pyproject.rs
aleksrutins: Correct me if I'm wrong, but I think that in this case it might actually make more sense to do the chaining:
1. It only needs access to a few of the fields in the file, so deserializing the entire data structure might actually add extra overhead.
2. There are fallbacks for each field, which seems like the sort of thing that chaining (specifically with `.or_else`) is useful for.

Plus, it's not nearly as painful with a macro.

src/providers/python.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    fn suggested_start_command(&self, app: &App, _env: &Environment) -> Result<Option<String>> {
+        if app.includes_file("pyproject.toml") {
+            if let Ok(meta) = self.parse_pyproject(app) {
+                if let Some(entry_point) = meta.entry_point {
+                    return match entry_point {
+                        EntryPoint::Command(cmd) => Ok(Some(cmd)),
+                        EntryPoint::Module(module) => {
+                            Ok(Some(format!("python -m {}", module)))
```
coffee-cup, line 47: I think these should `return`
```suggestion
                        EntryPoint::Command(cmd) => return Ok(Some(cmd)),
                        EntryPoint::Module(module) => {
                            return Ok(Some(format!("python -m {}", module)))
```

aleksrutins: It's already returning the result of the `match` expression, so I'm not sure that that's really necessary.

coffee-cup: Oh nice, I missed that

coffee-cup (approved):

PR #19, opened 2022-04-04
Add Deno provider

gschier:
This PR adds support for [Deno](https://deno.land/). Since Deno projects don't have any specific files we can use for detection, this PR adds a new `find_match` fn which will scan a glob of files and match a regular expression against the content for Deno-style imports.

_Note: this is my first experience with Rust so let me know if the code is trash!_

### DIFF
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -63,8 +63,7 @@ impl App {
 
             if let Some(p) = path_buf.to_str() {
                 let f = self.read_file(p)?;
-                let matches = re.find(f.as_str());
-                if matches.is_some() {
+                if re.find(f.as_str()).is_some() {
                     return Ok(true);
                 }
             }
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -183,11 +183,16 @@ mod tests {
     fn test_find_files() -> Result<()> {
         let app = App::new("./examples/monorepo")?;
         let m = app.find_files("**/*.tsx").unwrap();
+        let dir = env::current_dir().unwrap();
         assert_eq!(
             m,
             vec![
-                "packages/client/pages/_app.tsx",
-                "packages/client/pages/index.tsx"
+                dir.join("examples/monorepo/packages/client/pages/_app.tsx")
+                    .to_str()
+                    .unwrap(),
+                dir.join("examples/monorepo/packages/client/pages/index.tsx")
+                    .to_str()
+                    .unwrap(),
             ]
         );
         Ok(())
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -1,7 +1,7 @@
 use std::path::Path;
 use std::{env, fs, path::PathBuf};
 
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use glob::glob;
 use regex::Regex;
 use serde::de::DeserializeOwned;
@@ -93,7 +93,7 @@ impl App {
     pub fn strip_source_path(&self, abs: &str) -> Result<String> {
         let source_str = match self.source.to_str() {
             Some(s) => s,
-            None => return Err(anyhow::Error::msg("Failed to parse source path")),
+            None => bail!("Failed to parse source path"),
         };
         let abs_path = Path::new(abs);
 
@@ -106,7 +106,7 @@ impl App {
         // Convert path to string
         return match stripped.to_str() {
             Some(v) => Ok(v.to_string()),
-            None => Err(anyhow::Error::msg("Failed to convert path to string")),
+            None => bail!("Failed to convert path to string"),
         };
     }
 }
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -39,8 +39,7 @@ impl App {
 
         let relative_paths = glob(pattern_str)?
             .filter_map(|p| p.ok()) // Remove bad ones
-            .filter_map(|p| self.strip_source_path(p).ok()) // Make relative
-            .filter_map(|p| p.to_str().map(|p| p.to_string()))
+            .filter_map(|p| p.to_str().map(|p| p.to_string())) // convert to string
             .collect();
 
         Ok(relative_paths)
@@ -92,7 +91,7 @@ impl App {
         Ok(toml_file)
     }
 
-    fn strip_source_path(&self, abs: PathBuf) -> Result<PathBuf> {
+    pub fn strip_source_path(&self, abs: &str) -> Result<String> {
         let source_str = match self.source.to_str() {
             Some(s) => s,
             None => return Err(anyhow::Error::msg("Failed to parse source path")),
@@ -100,11 +99,11 @@ impl App {
 
         // Strip source path from absolute path
         let stripped = match abs.strip_prefix(source_str) {
-            Ok(p) => p.to_path_buf(),
-            Err(_e) => abs,
+            Some(p) => p,
+            None => abs,
         };
 
-        Ok(stripped)
+        Ok(stripped.to_string())
     }
 }
 
@@ -172,6 +171,48 @@ mod tests {
                 .unwrap(),
             "rocket"
         );
+        Ok(())
+    }
+
+    #[test]
+    fn test_find_files() -> Result<()> {
+        let app = App::new("./examples/monorepo")?;
+        let m = app.find_files("**/*.tsx").unwrap();
+        assert_eq!(
+            m,
+            vec![
+                "packages/client/pages/_app.tsx",
+                "packages/client/pages/index.tsx"
+            ]
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_find_match() -> Result<()> {
+        let app = App::new("./examples/monorepo")?;
+        let re = Regex::new(r"className")?;
+        let m = app.find_match(&re, "**/*.tsx").unwrap();
+        assert!(m);
+        Ok(())
+    }
+
+    #[test]
+    fn test_strip_source_path() -> Result<()> {
+        let app = App::new("./examples/npm")?;
+        let path_to_strip = app.source.join("foo/bar.txt");
+        assert_eq!(
+            app.strip_source_path(path_to_strip.to_str().unwrap())
+                .unwrap(),
+            "foo/bar.txt"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_strip_source_path_no_source_prefix() -> Result<()> {
+        let app = App::new("./examples/npm")?;
+        assert_eq!(app.strip_source_path("no/prefix.txt")?, "no/prefix.txt");
         Ok(())
     }
 }
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::{env, fs, path::PathBuf};
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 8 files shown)


Milo123459:
This PR could also check for the `Deno.` API.

Quick note, HTTP imports are getting added in Node.

src/providers/deno.rs
```diff
@@ -0,0 +1,37 @@
...
+        "deno"
+    }
+
+    fn detect(&self, app: &App) -> Result<bool> {
+        if !app.includes_file("src/index.ts") {
+            false;
+        }
+
+        let re = Regex::new(r##"(?m)^import .+ from "https://deno.land/[^"]+\.ts";?$"##).unwrap();
+        Ok(app.find_match(&re, "**/*.ts"))
```
Milo123459, line 19:
```suggestion
        Ok(app.find_match(&re, "**/*{.ts,.js}"))
```

src/providers/deno.rs
```diff
@@ -0,0 +1,37 @@
...
+    fn install_cmd(&self, _app: &App) -> Result<Option<String>> {
+        Ok(None)
+    }
+
+    fn suggested_build_cmd(&self, _app: &App) -> Result<Option<String>> {
+        Ok(None)
+    }
+
+    fn suggested_start_command(&self, _app: &App) -> Result<Option<String>> {
+        Ok(Some("deno run src/index.ts".to_string()))
```
Milo123459, line 35: Is this going to handle the explicit flags Deno needs, like `--allow-net`?

src/providers/deno.rs
gschier: Oh ya, nice catch!

src/nixpacks/app.rs
```diff
@@ -32,6 +35,22 @@ impl App {
         Ok(contents)
     }
 
+    pub fn find_match(&self, re: &Regex, pattern: &str) -> bool {
```
Milo123459, line 38: Just for documenting, I've created a little discussion over on the Discord about potentially making this function faster. [Link](https://canary.discord.com/channels/713503345364697088/960605685920452669/960669111279550484)

src/nixpacks/app.rs
```diff
@@ -32,6 +35,22 @@ impl App {
...
+        for entry in entries {
+            let path_buf = fs::canonicalize(entry.unwrap()).unwrap();
+            let f = self.read_file(path_buf.to_str().unwrap()).unwrap();
+            if re.find(f.as_str()).is_some() {
+                matched = true;
+                break;
+            }
+        }
+        matched
+    }
```
coffee-cup, line 52: Instead of unwrapping here I think the the `find_match` function should return a `Result`. Then you can `?` instead of unwrap the function calls.

src/providers/deno.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct DenoProvider {}
+
+impl Provider for DenoProvider {
+    fn name(&self) -> &str {
+        "deno"
+    }
+
+    fn detect(&self, app: &App) -> Result<bool> {
+        if !app.includes_file("src/index.ts") {
+            false;
```
coffee-cup, line 15:
```suggestion
            return false;
```

src/providers/deno.rs
coffee-cup: Yeah good point. I think it should allow-net by default. We can eventually use environment variables to control addition Deno behaviour.

src/nixpacks/app.rs
```diff
@@ -55,10 +74,11 @@ impl App {
 mod tests {
     use std::collections::HashMap;
 
-    use super::*;
     use serde::{Deserialize, Serialize};
     use serde_json::{Map, Value};
 
+    use super::*;
+
```
Milo123459, line 81: I don't think this changes anything, so perhaps this should be reverted to how it was?

src/providers/deno.rs
gschier: It does not. I wonder if we should run with `-A` like this [Heroku Deno builpack](https://github.com/chibat/heroku-buildpack-deno/blob/master/bin/compile) does since we don't know what permissions the user will actually want.

src/providers/deno.rs
```diff
@@ -0,0 +1,37 @@
...
+        }
```
Milo123459, line 16: This line actually does nothing, so we can remove it.

```suggestion
```

src/providers/deno.rs
gschier: `return Ok(false)` right?

src/providers/deno.rs
Milo123459: I mean, granting -A is generally dangerous, but as I don't see a better way of doing this, that's probably what is going to have to be done.

src/nixpacks/app.rs
Milo123459: For consistency, I think it should keep it as a bool because things like `includes_file` do it like this.

src/providers/deno.rs
gschier: It's because the run command references `src/index.ts` so I wanted to ensure it existed. We should probably make this. It should probably, instead, try to find either `index.ts` or `src/index.ts` and error if neither exist?

src/providers/deno.rs
Milo123459: FWIW, `src/index{.ts,.js}` isn't always the entry-point. The line actually does nothing and doesn't need to do anything because of the code below. This could just be completely removed, as per my review below.

src/providers/deno.rs
Milo123459: Is index.ts / index.js *always* the entry point, or can this be defined somewhere? Is the src directory needed as well?

src/nixpacks/app.rs
coffee-cup: Then we should make `includes_file` return a Result. A Result is much more useful since we can handle the error further up. Most other places in the providers/builder deal with results.

src/providers/deno.rs
Milo123459: Oh, actually it should just make sure that in the project there is either a JS or TS file. That can be done with glob, or perhaps introducing another function, `includes_glob`.

src/nixpacks/app.rs
Milo123459: There isn't actually a point to making `includes_file` return a `Result` because we manually check if it is ok with`is_ok`.

src/providers/deno.rs
gschier: Ya, will try something like that. Basically, run the first file that matches `**/index.{ts,js}`

coffee-cup: I think the problem is that Deno projects don't have a `package.json` file to inspect. For situations like this I don't think it is bad to go with a sane default. Sure it may not _always_ be the case, but it is likely. And users will be able to override the start command if they want. If this needs to be refined in the future then it can be. But one of the goals with this project is sane defaults with easy customization.

With that said, we could also check for `./index.ts` (no src) too since that is also pretty likely.

coffee-cup: > Ya, will try something like that. Basically, run the first file that matches `**/index.{ts,js}`

Oh I missed this in my reply. This is good I think

src/nixpacks/app.rs
coffee-cup: Sure, but your point above was about consistency. We shouldn't unwrap here and if we want everything to be consistent then we can return a Result everywhere. Not the other way around.

However, I think it is fine to sometimes return a Result and sometimes not. But unwrapping is akin to `!` in TypeScript, which is to be avoided.

src/providers/deno.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    fn pkgs(&self, _app: &App) -> Vec<Pkg> {
+        vec![Pkg::new("pkgs.stdenv"), Pkg::new("pkgs.deno")]
+    }
+
+    fn install_cmd(&self, _app: &App) -> Result<Option<String>> {
+        Ok(None)
+    }
+
+    fn suggested_build_cmd(&self, _app: &App) -> Result<Option<String>> {
```
Milo123459, line 30: In a later PR, support for `deno compile` could be added.

src/nixpacks/app.rs
Milo123459: Okay, I agree with you. However, `includes_file` shouldn't. We would be hardcoding an Ok() in `includes_file`.

src/providers/deno.rs
Milo123459: Actually, how does the heroku buildpack detect it?

src/nixpacks/app.rs
```diff
@@ -26,12 +30,48 @@ impl App {
...
+    pub fn find_files(&self, pattern: &str) -> Result<Vec<String>> {
+        let full_pattern = self.source.join(pattern);
+
+        let pattern_str = match full_pattern.to_str() {
+            Some(s) => s,
+            None => return Ok(Vec::new()),
+        };
+
+        let relative_paths = glob(pattern_str)?
+            .filter_map(|p| p.ok()) // Remove bad ones
```
coffee-cup, line 42: Should this be a filter map or just a filter?

src/nixpacks/app.rs
```diff
@@ -49,6 +89,26 @@ impl App {
...
+        // Strip source path from absolute path
+        let stripped = match abs_path.strip_prefix(source_str) {
+            Ok(p) => p,
+            Err(_e) => abs_path,
+        };
+
+        // Convert path to string
+        return match stripped.to_str() {
+            Some(v) => Ok(v.to_string()),
+            None => Err(anyhow::Error::msg("Failed to convert path to string")),
```
coffee-cup, line 109: for future ref can use the [`bail!`](https://docs.rs/anyhow/1.0.11/anyhow/macro.bail.html) macro

src/nixpacks/app.rs
```diff
@@ -117,4 +177,51 @@ mod tests {
...
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_strip_source_path_no_source_prefix() -> Result<()> {
+        let app = App::new("./examples/npm")?;
+        assert_eq!(app.strip_source_path("no/prefix.txt")?, "no/prefix.txt");
+        Ok(())
+    }
```
coffee-cup, line 226: I might be missing something, but what is the difference between these two tests?

coffee-cup (approved):
Looks great!

src/providers/deno.rs
gschier: Would compile actually gain us anything here? My understanding is that compiling simply wraps all files in an executable with the Deno runtime.

src/nixpacks/app.rs
Milo123459: Looks to be a filter. The function is being used wrong here so it makes sense.

gschier: `filter_map` is correct. It needs to map in order to convert the `Option` from `p.ok()` into a `PathBuf`

src/nixpacks/app.rs
gschier: Lol, one of them uses a path that's joined onto `app.source`. I wanted to test this because `PathBuf.strip_prefix` errors if the path does not contain the prefix.

src/nixpacks/app.rs
Milo123459: Ah, thought it was `is_ok`. My bad! 😅

src/providers/deno.rs
```diff
@@ -0,0 +1,44 @@
...
+use regex::Regex;
+
+pub struct DenoProvider {}
+
+impl Provider for DenoProvider {
+    fn name(&self) -> &str {
+        "deno"
+    }
+
+    fn detect(&self, app: &App) -> Result<bool> {
```
aleksrutins, line 13: I'm a bit late to the conversation, but I'd suggest also checking for `deno.json`, as that's a pretty definitive way to tell that it's Deno. Also, Deno supports import maps, which I'm finding in my projects are a handy way to pin dependencies, so a Deno app won't necessarily have any files that match that regex.

Issue #24, opened 2022-04-05
Environment variable support


coffee-cup:
Much like docker and pack, you should be able to provide environment variables that will be available to the build.

I am thinking they can be supplied with the flag

```
--env "NAME=value" EXTERNAL
```

In the NAME=value case, the variable NAME is set to "value". In the second case, the variable EXTERNAL is provided from the calling environment.

Milo123459:
Perhaps also specify a `.env` file that could be loaded?

PR #35, opened 2022-04-06
Environment variable support

coffee-cup:
Closes https://github.com/railwayapp/nixpacks/issues/24

This PR introduces an `Environment` abstraction that the app builder and providers can use to read environment variables. Eventually, we can extend the environment with more information (such as host os, time, etc).

- Add a new `--env` flag that can be used to pass environment variables to the build/plan phase. Variables are provided in the same way they are provided to the buildpack `pack` CLI. For example `--env "HELLO=world" "FOO"`. In this case `HELLO` will have the value "world" and the value for `FOO` will be pulled from the host environment.
- Add an `Environment` struct that abstracts setting and getting of environment variables
- Store the variables on the build plan
- Pass the environment to all provider methods
- New provider implementable method, `get_environment_variables` that providers can use to set default environment variable values. 
- The NPM/Yarn providers set the `NODE_ENV` and `NPM_CONFIG_PRODUCTION` variables.

image


src/providers/npm.rs
```diff
@@ -12,19 +15,19 @@ impl Provider for NpmProvider {
         "node"
     }
 
-    fn detect(&self, app: &App) -> Result<bool> {
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
```
JakeCooper, line 18: Should the environment exist on the app itself?

PR #36, opened 2022-04-06
install npm/yarn dependencies from lockfile

coffee-cup:
- Install NPM deps with `npm ci`
- Install Yarn deps with `yarn install --frozen-lockfile`

In both cases this will install deps from the lockfile directly and will not update the lockfile with changes.

### DIFF
--- a/src/providers/npm.rs
+++ b/src/providers/npm.rs
@@ -24,7 +24,7 @@ impl Provider for NpmProvider {
     }
 
     fn install_cmd(&self, _app: &App, _env: &Environment) -> Result<Option<String>> {
-        Ok(Some("npm install".to_string()))
+        Ok(Some("npm ci".to_string()))
     }
 
     fn suggested_build_cmd(&self, app: &App, _env: &Environment) -> Result<Option<String>> {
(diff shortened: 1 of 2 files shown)


JakeCooper (approved):

devtomio:
iirc `--frozen-lockfile` is yarn 1 only, yarn 2 and above use `--immutable`

coffee-cup:
> iirc `--frozen-lockfile` is yarn 1 only, yarn 2 and above use `--immutable`

Only Yarn 1 is supported at the moment. A second provider will most likely have to be added for Yarn 2 (or we will have to detect and install the proper Nix packages in the current Yarn provider)

coffee-cup:
In fact, I can't seem to find a Yarn 2 package on [Nix packages](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=yarn) so we will probably have to provide that derivation ourselves.

devtomio:
> In fact, I can't seem to find a Yarn 2 package on [Nix packages](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=yarn) so we will probably have to provide that derivation ourselves.

one would just need to detect the `.yarnrc.yml` to detect if an app is running on yarn >2

PR #37, opened 2022-04-07
Node package.json engines parsing and package overriding

coffee-cup:
This is the first step in providing custom packages versions and package overrides.

In this PR
- Allow overriding of sub Nix packages `Pkg`
- Share some logic between the yarn and npm providers
- Parse the `engines.node` field of `package.json` files to find a best matching Nix package. Currently versions 10, 12, 14, 16, and 17 are available.
- basic node version parsing
- pkg overriding and tests

In a follow up PR I would like to add
- `.nvmrc` (and other forms of) Node version parsing 
- More robust engines field parsing
- Override rust version


devtomio:
shouldn't this also include `.node-version` parsing?

devtomio:
also it should check for the `NODE_VERSION` environment variable

JakeCooper (approved):

coffee-cup:
> shouldn't this also include `.node-version` parsing?
> also it should check for the `NODE_VERSION` environment variable

Yes both of these will come in a follow up PR (see description). I've created an issue to track these https://github.com/railwayapp/nixpacks/issues/38

PR #53, opened 2022-04-27
Debian base image

coffee-cup:
Previously the base image was [nixos/nix](https://hub.docker.com/r/nixos/nix/dockerfile). This is based on the minimal alpine image. When using nixpacks to build and deploy some real world apps (specifically prisma), there were some issues because alpine is based on musl libc instead of glibc and prisma was failing to find binaries. Alpine linux is great, but it does come with some quirks and it is not uncommon to require some manual user intervention to get binaries installed and running properly. Debian based images are much more common and supported.With the zero-config nature of Nixpacks I think it makes sense to move to debian based images.

This PR changes the base image to be `debian:buster-slim` (Debian 11) and installs Nix manually. The final image size (for a yarn app) is increase by ~20%.

I will follow up with another PR that focuses on decreasing the final image size and separating build and runtime images.


JakeCooper (commented):
Should we not have this be like, something that people using Prisma can add for their deps?

JakeCooper:
I'm fine either way FWIW, I'm just wondering where we draw the line

PR #65, opened 2022-05-06
Config through environment variables

coffee-cup:
Allow specifying config through `NIXPACKS_*` environment variables. This is useful when using Nixpacks on Railway. The knobs that can be tweaked are

- Nix packages to install
- Build command
- Start command
- Node version
- Rust version

Other properties (e.g. Python version) will comes soon once they are setup properly.

### DIFF
--- a/src/nixpacks/environment.rs
+++ b/src/nixpacks/environment.rs
@@ -16,6 +16,10 @@ impl Environment {
         self.variables.get(name)
     }
 
+    pub fn get_config_variable(&self, name: &str) -> Option<&String> {
+        self.get_variable(format!("NIXPACKS_{}", name).as_str())
+    }
+
     pub fn set_variable(&mut self, name: String, value: String) {
         self.variables.insert(name, value);
     }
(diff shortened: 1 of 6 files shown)


JakeCooper:
Any reason we need to prefix it?

coffee-cup:
No real reason. I just wanted a clear separation between variables to configure nixpacks itself vs variables for the user app

JakeCooper:
Ya that's fair. I wonder if we could come up with something a little less, domain specific?

RUNTIME_ENV_*? Cause I could see us wanting at some point a BUILD_ENV_*

JakeCooper:
Oh wait I've flipped them. This is for ones that are being provided to Nixpacks themselves

Sorry

JakeCooper (approved):

coffee-cup:
> Oh wait I've flipped them. This is for ones that are being provided to Nixpacks themselves

Yeah this is for configuring things in Nixpacks itself (start, build cmds, etc)

I have also been thinking we need a separation between build and runtime environment variables. But I think we need a bit more data on how Nixpacks is used on real apps first since the edge cases can get complicated (or I could just be over thinking it). Like providers can specify variables that should be used at build and/or runtime and users can override these variables or specify their own custom build/runtime variables. Right now all user/provider variables are just grouped together which makes overriding and using them simple. But I can see us having more complex variable handling in the future.

Issue #71, opened 2022-05-07
If package manifest isn't found, don't do a clean install


Milo123459:
This is causing issues because some people don't commit `yarn.lock` & `package-lock.json`. We need to remove `--frozen-lockfile` in Yarn if `yarn.lock` doesn't exist, and we need to do `npm i` instead of `npm ci` if `package-lock.json` doesn't exist.

coffee-cup:
Fixed in https://github.com/railwayapp/nixpacks/pull/89

PR #72, opened 2022-05-07
fix(npm): fallback to `npm install`

wyzlle:
Falls back to `npm install` if `npm ci` fails.

### DIFF
--- a/src/providers/npm.rs
+++ b/src/providers/npm.rs
@@ -42,7 +42,7 @@ impl Provider for NpmProvider {
     }
 
     fn install(&self, _app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
-        let install_phase = InstallPhase::new("npm ci".to_string());
+        let install_phase = InstallPhase::new("npm ci || npm install".to_string());
         Ok(Some(install_phase))
     }
 


Milo123459:
Fixes #71

src/providers/npm.rs
```diff
@@ -42,7 +42,7 @@ impl Provider for NpmProvider {
     }
 
     fn install(&self, _app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
-        let install_phase = InstallPhase::new("npm ci".to_string());
+        let install_phase = InstallPhase::new("npm ci || npm install".to_string());
```
Milo123459, line 45: I think checking if `package-lock.json` exists and then running `npm ci` if it does is a better idea.

Milo123459:
Could you make this work for Yarn as well?

wyzlle:
> Could you make this work for Yarn as well?

There's no way to find which package manager to use if lockfile isn't found.

Milo123459:
> > Could you make this work for Yarn as well?
> 
> There's no way to find which package manager to use if lockfile isn't found.

There is. Can't remember how though. It's something in the package.json though.

aleksrutins:
There is actually now, through the new package.json `packageManager` field: https://nodejs.org/api/corepack.html#configuring-a-package

Milo123459:
I'll implement support for that in a later PR. Other then that and my review, LGTM.

src/providers/npm.rs
wyzlle: If dependencies in the package lock do not match those in package.json npm ci fails which isn't ideal. Also, it's not just `package-lock.json` ci works with `npm-shrinkwrap.json` too

wyzlle:
> There is actually now, through the new package.json `packageManager` field: https://nodejs.org/api/corepack.html#configuring-a-package

`Corepack` is an experimental tool. Idt npm/yarn/pnpm adds that field yet(Correct me if I am wrong). Would be better if we make this change once corepack becomes stable.

aleksrutins:
Yarn and pnpm do actually add that field already.

wyzlle:
> Yarn and pnpm do actually add that field already.

Oh, can implement then thanks for clarifying. Will update this PR in a bit.

src/providers/npm.rs
coffee-cup: I agree with Milo. The entire point of the lockfile is to set exact versions of a dependency. If the versions in the lockfile do not match those in package.json then building with Nixpacks **should** fail.

However, if you don't have a lockfile specified then that is okay and we can fallback to just running `npm install`.

wyzlle: Buildpacks don't do that and I have seen repos (including  [railway starters](https://github.com/railwayapp-starters/ghost/commit/9b31609b8fb503657716a550d4d51dc11aef8bf4)) where people just update package.json temporarily to update to a new version of that dependency. Won't be backward compatible if we make this change. Up to you/team I don't mind adding that.

coffee-cup: I think the Paketo buildpacks also install with `npm ci` or `yarn --frozen-lockfile`. IMO when updating a version the lockfile should also be regenerated.

In this case I think we should respect the lockfile contents.

wyzlle: Sure, will update PR accordingly.

PR #87, opened 2022-05-09
Haskell provider

### LINKED ISSUE #81: Haskell provider

aleksrutins:
Fixes #81


src/lib.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
     },
 };
 use anyhow::{bail, Result};
-use providers::Provider;
+use providers::{haskell::HaskellStackProvider, Provider};
```
coffee-cup, line 14: nit: Add this import to the above list (similar to how other providers are imported)

src/nixpacks/mod.rs
```diff
@@ -497,7 +497,7 @@ impl<'a> AppBuilder<'a> {
           FROM {base_image}
 
           RUN apt-get update && apt-get -y upgrade \\
-            && apt-get install --no-install-recommends -y locales curl xz-utils ca-certificates openssl \\
+            && apt-get install --no-install-recommends -y sudo locales curl xz-utils ca-certificates openssl \\
```
coffee-cup, line 500: What is the reasoning for installing `sudo` here? The apt packages below should be able to be installed without it

src/providers/haskell.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    fn setup(
+        &self,
+        _app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::phase::SetupPhase>> {
+        Ok(Some(SetupPhase::new(vec![
+            Pkg::new("stack"),
+            Pkg::new("apt"),
```
coffee-cup, line 34: Why is `apt` needed?

src/providers/haskell.rs
```diff
@@ -0,0 +1,76 @@
...
+        ])))
+    }
+
+    fn install(
+        &self,
+        _app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::phase::InstallPhase>> {
+        Ok(Some(InstallPhase::new("sudo apt-get update && sudo apt-get install -y libgmp-dev gcc binutils make && stack setup".to_string())))
```
coffee-cup, line 43: Can these packages be installed with Nix instead? `gcc` and `make` should be able to. What are `libgmp-dev` and `binutils` for?

src/providers/haskell.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn name(&self) -> &str {
+        "haskell_stack"
+    }
+
+    fn detect(
+        &self,
+        app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<bool> {
+        Ok(app.includes_file("package.yaml"))
```
coffee-cup, line 24: I think this should check for something more than a single yaml file to check if the app is Haskell based. What about checking for `.cabal` or `.hs` files?

src/nixpacks/mod.rs
aleksrutins: `sudo` is needed to install the apt packages in the provider itself, as the setup phase runs as a separate build user instead of directly as `root`.

src/providers/haskell.rs
aleksrutins: `apt` is needed in case we’re using a base image other than Debian at some point in the future.

src/providers/haskell.rs
aleksrutins: Nix doesn’t package any development libraries/headers. `libgmp-dev` is a floating-point math library needed by the Haskell compiler. `binutils` has things like `ar` that are needed to build libraries. `make` could probably be installed through Nix, but `gcc` needs to be installed using the same package manager as any libraries you’re trying to use, so that it can find them. Since `apt` is the only way to get development libraries, `gcc` needs to be installed through it too.

src/providers/haskell.rs
aleksrutins: I was originally thinking that there could be another provider for directly Cabal-based projects. This is specifically for Stack projects, and Stack uses `package.yaml`. I was originally also looking for `stack.yaml`, but it looks like that’s not actually necessarily always going to be there.

coffee-cup: That makes sense. I think checking for the existing of a `.hs` file would also be useful here too. Only because `package.yaml` is not Haskell specific could be used for another language.

src/providers/haskell.rs
coffee-cup: Ah I see. I thinkkk this is okay. I'm hesitant to start having providers download external dependencies outside of Nix since we lose the ability to pin to a specific version of the Nix store. I think this is fine for now but we should be cautious about encouraging this moving forward.

If more than one provider can be matched at a time (future plans) then it would be nicer if the Haskell provider could depend on the AptProvider and specify the packages that way. Then at least the behaviour is consistent and more controller than just installing w/e packages in the install phase.

src/providers/haskell.rs
```diff
@@ -0,0 +1,76 @@
...
+    phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
+};
+
+use super::Provider;
+
+pub struct HaskellStackProvider {}
+
+impl Provider for HaskellStackProvider {
+    fn name(&self) -> &str {
+        "haskell_stack"
```
Milo123459, line 16: Shouldn't this just be `haskell`?

src/lib.rs
aleksrutins: Fixed!

src/providers/haskell.rs
aleksrutins: Fixed! It now looks for `package.yaml`, plus anything with a `.hs` extension anywhere in the source tree.

coffee-cup:
Can you please fix the conflicts. I can merge once complete

aleksrutins:
Oh gosh, rebasing broke everything again, didn't it...

coffee-cup (approved):

PR #89, opened 2022-05-10
Merge npm and yarn providers into node provider

wyzlle:


src/providers/node.rs
```diff
@@ -36,26 +36,47 @@ impl Provider for NpmProvider {
...
+            if node_pkg.name != *DEFAULT_NODE_PKG_NAME {
+                yarn_pkg = yarn_pkg.set_override("nodejs", node_pkg.name.as_str());
+            }
+            return Ok(Some(SetupPhase::new(vec![node_pkg, yarn_pkg])));
+        }
 
         Ok(Some(SetupPhase::new(vec![node_pkg])))
     }
 
     fn install(&self, _app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
```
coffee-cup, line 52: Nit: change `_app` to `app`. Prefixed underscore should only be used if variable is unused

```suggestion
    fn install(&self, app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
```

coffee-cup (approved):
Nice! One small suggestion but looks good otherwise

src/providers/node.rs
wyzlle: change `_app` to `app`

coffee-cup:
Also in the future can you please add a PR description. Even something small is useful

wyzlle:
> Also in the future can you please add a PR description. Even something small is useful

Sure, title was pretty self explanatory that's why I didn't add any description.

Issue #69, opened 2022-05-07
PNPM provider


Milo123459:

wyzlle:
Working on this right now.

PR #91, opened 2022-05-10
Add pnpm provider

wyzlle:
Fixes https://github.com/railwayapp/nixpacks/issues/69

### DIFF
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -58,27 +58,21 @@ impl Provider for NodeProvider {
     }
 
     fn install(&self, app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
-        let install_cmd: InstallPhase;
+        let mut install_cmd = "npm i";
         if NodeProvider::get_package_manager(app)? == "pnpm" {
-            install_cmd = InstallPhase::new("pnpm i --frozen-lockfile".to_string());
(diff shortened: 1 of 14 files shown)


src/providers/node.rs
```diff
@@ -50,20 +58,28 @@ impl Provider for NodeProvider {
...
+            install_cmd = InstallPhase::new("yarn install --frozen-lockfile".to_string());
+            install_cmd.add_file_dependency("yarn.lock".to_string());
+        } else if app.includes_file("package-lock.json") {
+            install_cmd = InstallPhase::new("npm ci".to_string());
+            install_cmd.add_file_dependency("package-lock.json".to_string());
+        } else {
+            install_cmd = InstallPhase::new("npm i".to_string());
         }
-        Ok(Some(InstallPhase::new("npm install".to_string())))
+        install_cmd.add_file_dependency("package.json".to_string());
```
coffee-cup, line 74: These file dependencies used to be here but I removed them since it breaks some apps since some apps run post-install scripts from something like `bin/scripts`. It is unfortunate but we might be able to add it back in the future with some more advanced detection.

For now I think it should be removed :(

wyzlle: Cool, removed.

coffee-cup (approved):
Looks great! Thanks ✨

PR #92, opened 2022-05-11
Add support for yarn2+

### LINKED ISSUE #73: Yarn 2 support

wyzlle:
Fixes #73


tests/generate_plan_tests.rs
```diff
@@ -93,6 +93,23 @@ fn test_yarn() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_yarn_berry() -> Result<()> {
```
coffee-cup, line 97: Could you please add a docker run test for this too. Also for the PNPM stuff would also be useful. Should just a few lines for each test.

coffee-cup (approved):

tests/generate_plan_tests.rs
wyzlle: Oh yeah, I forgot about that.

PR #99, opened 2022-05-11
copy files manually rather than invoking cp

### LINKED ISSUE #98: `build` fails on Windows, due to `cp` not existing
The `cp` shell command is not available on Windows, therefore building fails when the program attempts to copy source files into a temporary directory.

nebulatgs:
Fixes the `build` command on Windows by manually copying files rather than using UNIX shell commands.

Fixes #98

### DIFF
--- a/src/nixpacks/mod.rs
+++ b/src/nixpacks/mod.rs
@@ -2,12 +2,13 @@ use anyhow::{bail, Context, Ok, Result};
 use indoc::formatdoc;
 use std::{
     fs::{self, File},
-    io::Write,
+    io::{self, Write},
     path::PathBuf,
     process::Command,
 };
 use tempdir::TempDir;
 use uuid::Uuid;
+use walkdir::WalkDir;
 pub mod app;
 pub mod environment;
 pub mod logger;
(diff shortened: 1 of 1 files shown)


src/nixpacks/mod.rs
```diff
@@ -138,27 +139,43 @@ impl<'a> AppBuilder<'a> {
...
-        if !copy_result.success() {
-            bail!("Copy failed")
+        let in_dir = PathBuf::from(source);
+
+        let walker = WalkDir::new(&in_dir).follow_links(true);
+        for entry in walker {
+            let entry = entry?;
+
+            let from = entry.path();
+            dbg!(&from, &dir, &in_dir);
```
coffee-cup, line 160: 🔫  (remove this line)

src/nixpacks/mod.rs
```diff
@@ -138,27 +139,43 @@ impl<'a> AppBuilder<'a> {
...
-            .arg("-a")
-            .arg(format!("{}/.", source))
-            .arg(dir.clone())
-            .spawn()?;
-        let copy_result = copy_cmd.wait().context("Copying app source to tmp dir")?;
-        if !copy_result.success() {
-            bail!("Copy failed")
+        let in_dir = PathBuf::from(source);
+
+        let walker = WalkDir::new(&in_dir).follow_links(true);
```
coffee-cup, line 155: I think this should be moved out into a function

src/nixpacks/mod.rs
nebulatgs: 😨

coffee-cup (approved):

PR #108, opened 2022-05-11
publish custom images to ghcr

wyzlle:
Builds & pushes `base/debian/Dockerfile` to `ghrc.io/railwayapp/nixpacks` as discussed [here](https://canary.discord.com/channels/713503345364697088/974023264143831060).

### DIFF
--- a/.github/workflows/publish_debian.yml
+++ b/.github/workflows/publish_debian.yml
@@ -1,0 +1,43 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'base/debian/*'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'base/debian/*'
+
+name: Publish Docker Images
+
+jobs:
+  debian:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Login to GitHub Container Registry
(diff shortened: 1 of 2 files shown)


base/debian/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM debian:bullseye-20220509-slim
```
nebulatgs, line 1: any reason to lock the version here?

wyzlle: Yeah, so nothing breaks in future. There's no downside to this but if you want I can remove that.

coffee-cup: I think locking it makes sense. Better to be exact about which version to use.

.github/workflows/publish_debian.yml
```diff
@@ -0,0 +1,43 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'base/debian/*'
+  pull_request:
```
coffee-cup, line 7: I don't think this should run on pull requests, just pushes to `main`. Otherwise a pull request into `main` will publish a new image.

wyzlle: It'll only push when code is commited to main. [click here](https://github.com/wyzlle/nixpacks/blob/main/.github/workflows/publish_debian.yml#L42)

coffee-cup: Ah I see, missed that.

coffee-cup (approved):

PR #109, opened 2022-05-12
Add support for more rust targets

devtomio:
This pull request adds support for more rust targets (e.g. `x86_64-unknown-freebsd`, `i686-pc-windows-msvc`, etc.). I've also adjusted the install script to work with that.


coffee-cup (approved):
Love the improvements to the install script!

Thanks for adding additional targets too 😄

PR #106, opened 2022-05-11
Minimal go images

coffee-cup:
- **Build plan update**: Base image is now specified in the setup phase
- **Provider API update**: Providers can specify a `run image` in the start phase. If this image is provided then we will add the following to the generated Dockerfile
```
FROM {run_image}
WORKDIR /app/
COPY --from=0 /app /app/
```

so that only the app source and any build artifacts will be present in the final build. This is useful for compiled langauges that can produce a single staticically linked binary.

- Compile Go with `CGO_ENABLED=0`

The result is that the go image produced for `examples/go-mod` drops from 1.15GB to 76MB

The run image can be overridden by setting the `NIXPACKS_RUN_IMAGE` environment variable.

PR #113, opened 2022-05-12
Change base image to `ghcr.io/railwayapp/nixpacks:debian`

wyzlle:
Changing base image will make builds faster and reproducible. Also, better log readability.


wyzlle:
Will reopen after #106 is merged.

coffee-cup:
I think we need to install `sudo` here for the Haskell tests to pass https://github.com/railwayapp/nixpacks/blob/main/base/debian/Dockerfile#L4

coffee-cup (approved):

PR #124, opened 2022-05-14
`npm run start` -> `npm start`

wyzlle:
Accordingly, start command will change for other package managers too.
`yarn run start` -> `yarn start`
`pnpm run start` -> `pnpm start`

### DIFF
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -123,7 +123,7 @@ impl NodeProvider {
 
     pub fn get_start_cmd(app: &App) -> Result<Option<String>> {
         if NodeProvider::has_script(app, "start")? {
-            return Ok(Some("npm run start".to_string()));
+            return Ok(Some("npm start".to_string()));
         }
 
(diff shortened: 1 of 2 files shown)


src/providers/node.rs
```diff
@@ -123,7 +123,7 @@ impl NodeProvider {
 
     pub fn get_start_cmd(app: &App) -> Result<Option<String>> {
         if NodeProvider::has_script(app, "start")? {
-            return Ok(Some("npm run start".to_string()));
+            return Ok(Some("npm start".to_string()));
```
Milo123459, line 126: Verbosity is better, I don't think this is a good change.

wyzlle: Previously it was `npm start`. I am fine either ways.

coffee-cup: Agree with @Milo123459 that verbosity is better. Would prefer to stick with `npm run ...` versions

PR #135, opened 2022-05-16
Fix go runtime image

coffee-cup:
Run go code in debian slim, not the nixpacks debian image

### DIFF
--- a/src/nixpacks/images.rs
+++ b/src/nixpacks/images.rs
@@ -1,0 +1,3 @@
+pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1652414952";
+
+pub static DEBIAN_SLIM_IMAGE: &str = "debian:bullseye-slim";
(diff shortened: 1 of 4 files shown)


src/nixpacks/images.rs
```diff
@@ -0,0 +1,3 @@
+pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1652414952";
```
Milo123459, line 1: I don't think we should be pinning it to a specific version.

coffee-cup: Why not? The base image is stored in the build plan and makes builds more reproducible. If it just uses the latest image then an update may break the build.

Milo123459: Whenever a new version of the image goes out, we have to manually change the source code. I don't think this is a good idea (perhaps we can make it auto-update?)

coffee-cup: Yeah thats true, but it means we are being explicit about the image. Alternatively we could fetch the latest when nixpacks is run and then use that in the build plan. But that can come in a separate PR.

coffee-cup: Going to merge this for now since I prefer being explicit and pinning the dep over always pulling the latest. At least when it comes to the build plan generated.

PR #136, opened 2022-05-16
Add make to base image

coffee-cup:
I've been going back and forth on this. Preferrably packages are installed through Nix, but in reality a lot of language and framework dependencies assume/expect the standard GCC/make toolchain to be available. This is the case for some Node, Python, and Ruby packages. For the reason I think adding `build-essential` to the default packages available on Debian makes sense.


coffee-cup:
Hmm quite a bit unfortunately. On second thought I am just going to add make since it isn't available (unless I am blind) on nixpkgs.

PR #137, opened 2022-05-16
don't require a "name" field in package.json

coffee-cup:
We should be able to build just fine without it

### DIFF
--- a/examples/node/package.json
+++ b/examples/node/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "node",
   "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


examples/node/package.json
```diff
@@ -1,5 +1,4 @@
 {
-  "name": "node",
```
Milo123459, line 2: Is there a real reason to remove it from the example?

coffee-cup: No, was just testing. I can revert

PR #140, opened 2022-05-16
Smaller final Rust image sizes

coffee-cup:
This PR builds Rust apps with the `{arch}-unknown-linux-musl` target by default. This creates a static binary which is then copied over to a fresh debian slim image resulting in a significantly smaller image size (~3GB -> 85MB).

Related discussion: https://github.com/railwayapp/nixpacks/discussions/139

image

_PR still a draft as I clean it up and make it configurable_

## Todo

- [ ] Detect parent arch and make rust target arch dynamic
- [ ] Allow opting out of musl builds with an env var
- [ ] Providers shouldn't need to know about the "app" directory
- [ ] Handle rust apps with a rust-toolchain file

### DIFF
--- a/src/nixpacks/mod.rs
+++ b/src/nixpacks/mod.rs
@@ -513,22 +513,28 @@ impl<'a> AppBuilder<'a> {
             .unwrap_or_else(|| "".to_string());
 
         // If we haven't yet copied over the entire app, do that before starting
-        let mut start_files: Vec<String> = Vec::new();
-        if build_phase.only_include_files.is_some() {
-            start_files.push(".".to_string());
-        }
+        let start_files: Vec<String> = start_phase
+            .only_include_files
+            .clone()
+            .unwrap_or_else(|| vec![".".to_string()]);
 
-        let run_image_setup = start_phase.run_image.map(|run_image| {
-            formatdoc! {"
+        let run_image_setup = match start_phase.run_image {
+            Some(run_image) => {
+                formatdoc! {"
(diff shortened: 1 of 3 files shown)


devtomio:
Does this work with crates that have libraries dynamically linked to them? For example, [`sodiumoxide`](https://crates.io/crates/sodiumoxide).

wyzlle:
> Could this approach be taken for all languages by default?

Yes, for compiled languages.

Milo123459:
> > Could this approach be taken for all languages by default?
> 
> No, only for compiled languages.

Maybe something to mark the provider as a compiled language, which we could then use this technique for all compiled languages?

coffee-cup:
> > > Could this approach be taken for all languages by default?
> > 
> > 
> > No, only for compiled languages.
> 
> Maybe something to mark the provider as a compiled language, which we could then use this technique for all compiled languages?

This technique is Rust specific. Other languages of different ways of compiling for a specific target

coffee-cup:
> Also, shouldn't we use `x86_64-unknown-linux-musl` instead of `aarch64-unknown-linux-musl`?

We should detect the host architecture and use that (see todos). I was just testing this on my arm machine locally so used aarch64.

> Does this work with crates that have libraries dynamically linked to them? For example, [`sodiumoxide`](https://crates.io/crates/sodiumoxide). 

Unfortunately, it won't. But there will be a way to disable it. Also by the target can be overridden by specifying a `rust-toolchain.toml` file.

PR #145, opened 2022-05-26
Flag to add additional labels to resulting image

### LINKED ISSUE #142: Add a way to specify container labels on the built images
It would be great to be able to specify one or more additional labels on the built images. These could be used to attach extra build metadata that can be inspected by other tooling.

coffee-cup:
New flag to add additional labels to the resulting Docker image. For example,

```
nixpacks build examples/XXX --label one two three
```

screenshot-2022-05-26-11 42 02

Closes #142

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,6 +49,7 @@ pub fn gen_plan(
         out_dir: None,
         plan_path: None,
         tags: Vec::new(),
+        labels: Vec::new(),
         quiet: false,
     };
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -72,6 +73,7 @@ pub fn build(
     plan_path: Option<String>,
     out_dir: Option<String>,
     tags: Vec<&str>,
+    labels: Vec<&str>,
```
Milo123459, line 76: Why are we using `&str` as opposed to `String`?

coffee-cup: No big reason. It is just the type that comes from clap's `values_of` https://github.com/railwayapp/nixpacks/blob/main/src/main.rs#L119

PR #150, opened 2022-05-30
add versioning support for golang provider

### LINKED ISSUE #148: Golang Version Provider
Golang Nixpack should support versioning, defined inside of go.mod

It should NOT use custom flagging like the Heroku one (e.g should parse the inline go version)

jozsefsallai:
Added support for retrieving the Go version number from `go.mod` (if exists). Also added an example and tests for the version parsing method, the docker run output, and plan generation.

Note: I've also updated `examples/go-mod`, as it was pointing to a Go version that's not available on nix unstable.

Closes #148

### DIFF
--- a/src/providers/go.rs
+++ b/src/providers/go.rs
@@ -5,11 +5,13 @@ use crate::nixpacks::{
     nix::Pkg,
     phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
 };
-use anyhow::Result;
+use anyhow::{bail, Result};
 
 pub struct GolangProvider {}
 
 pub const BINARY_NAME: &'static &str = &"out";
+const AVAILABLE_GO_VERSIONS: &[(&str, &str)] = &[("1.17", "go"), ("1.18", "go_1_18")];
+pub const DEFAULT_GO_PKG_NAME: &'static &str = &"go";
 
 impl Provider for GolangProvider {
     fn name(&self) -> &str {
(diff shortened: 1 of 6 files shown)


src/providers/go.rs
```diff
@@ -72,3 +77,79 @@ impl Provider for GolangProvider {
...
+        Ok(DEFAULT_GO_PKG_NAME.to_string())
+    }
+}
+
+fn version_number_to_pkg(version: &str) -> Result<Option<String>> {
+    let matched_version = AVAILABLE_GO_VERSIONS.iter().find(|(v, _)| v == &version);
+
+    match matched_version {
+        Some((_, pkg)) => Ok(Some(pkg.to_string())),
+        None => bail!("Go version {} is not available", version),
```
JakeCooper, line 113: I think if the version isn't specified we should default to the `go` package. Thoughts?

Milo123459: I agree, this should be how we handle it for all providers. If the version isn't found, default to the version on the nixos tag (like the node provider).

coffee-cup (approved):
Looks good (besides falling back to default go version if none found)! Thanks for adding this.

src/providers/go.rs
jozsefsallai: The node provider will actually error out if you specify a version in engines that's not in the list of supported versions (I tested the behavior myself). I wanted to preserve the same behavior as on node. But I can change this later.

wyzlle: There's a difference between specifying unsupported version and not specifying any version. Node provider falls back to [default version](https://github.com/railwayapp/nixpacks/blob/fcf36f29ab8455f10a4c546ea6c58a9f296c197c/src/providers/node.rs#L156) if version isn't specified.

PR #152, opened 2022-05-30
Add poetry support

### LINKED ISSUE #74: python-poetry provider

wyzlle:
Fixes #74

### DIFF
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -8,7 +8,7 @@ use crate::{
     chain,
     nixpacks::{
         app::App,
-        environment::Environment,
+        environment::{Environment, EnvironmentVariables},
         phase::{InstallPhase, SetupPhase, StartPhase},
     },
     Pkg,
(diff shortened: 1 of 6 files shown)


src/providers/python.rs
```diff
@@ -80,6 +91,19 @@ impl Provider for PythonProvider {
...
     }
+    fn environment_variables(
+        &self,
+        app: &App,
+        _env: &Environment,
+    ) -> Result<Option<EnvironmentVariables>> {
+        if app.includes_file("poetry.lock") {
+            return Ok(Some(EnvironmentVariables::from([(
+                "POETRY_VERSION".to_string(),
+                "1.1.13".to_string(),
```
Milo123459, line 102: Is this the latest version?

wyzlle: Yes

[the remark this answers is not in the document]
wyzlle: > Also, why is the version specified as an env var. Is it not specified in the requirements.txt file? Instead of using an env var I think we should just use a version by default and allow overriding the version with an env var.

There's no `requirements.txt` in projects that use poetry as a package manager.

coffee-cup (approved):

coffee-cup:
Thanks!

PR #154, opened 2022-05-30
Start working on Ruby support

### LINKED ISSUE #26: Ruby on Rails provider
* Provider for Ruby
* Should support Sinatra and Rails

Milo123459:
Fixes #26


src/providers/ruby.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn setup(&self, app: &App, _env: &Environment) -> Result<Option<SetupPhase>> {
+        let framework = self.detect_framework(app);
+        let mut packages = vec![Pkg::new("ruby")];
+        let needs_java = app.find_match(&Regex::new(r"jruby")?, "Gemfile.lock")?;
+        if needs_java {
+            packages.push(Pkg::new("java"));
+        }
+        match framework {
+            Framework::Rails => {
+                packages.append(&mut vec![Pkg::new("postgresql"), Pkg::new("nodejs")]);
```
coffee-cup, line 37: I think Rails applications have a `package.json` so we could get the node version from that. Would be something like `NodeProvider::get_nix_node_pkg`

coffee-cup (commented):
Can you please add a rails and vanilla example

src/providers/ruby.rs
Milo123459: From an example of a rails app I cannot see a package.json in root. Are you sure there is one?

wyzlle: Can use this https://github.com/cloudfoundry/ruby-buildpack/tree/master/fixtures for testing ruby provider.

All `rails5*` directores have `package.json`

coffee-cup: It exists in this Rails blog example https://github.com/railwayapp/starters/blob/master/examples/rails-blog/package.json

Not sure how node deps would be specified without one. Maybe it is optional?

Milo123459: I've added node version and packaging support. Most of this logic should probably be cleaned in the Node provider, I might do that later on.

Milo123459:
Something else to think about that is nice and fun:
There are about a thousand ways to start a rails app. Using rake, puma or just using bundle with rails. I guess I'm going to have to implement some sort of matching system for what build tool is used.

Milo123459:
When #156 is merged, I can rebase this PR and remove most of the logic for Node stuff being handled in the Ruby provider.

coffee-cup (commented):
This is coming along nicely! I have a few comments

- The vanilla ruby example fails to build since there is no Gemfile or .bundle directory

image

- There should be a rails example
  + Minor nit, but can the examples start with `ruby-*` (so like `ruby-vanilla` and `ruby-rails`). That way they are easier to see in the examples directory
- There should be generate plan and docker run tests for both ruby examples

src/providers/ruby.rs
```diff
@@ -0,0 +1,119 @@
...
+        _app: &App,
+        _env: &Environment,
+    ) -> Result<Option<EnvironmentVariables>> {
+        Ok(None)
+    }
+}
+
+impl RubyProvider {
+    fn detect_framework(&self, app: &App) -> Framework {
+        if app.includes_file("Rakefile") {
```
jaredcwhite, line 113: I don't think the presence of a `Rakefile` indicates a Rails app. I work on projects using the Bridgetown framework, Roda, etc. and they have Rakefiles. Probably need to look for a Rails initializer or something like that.

Milo123459: Are there any files that are specific to Rails?

JakeCooper: Should be able to just check the Gemfile for `gem 'rails'`

JakeCooper: I think I did a similar thing for Django

JakeCooper: Cause also pull the ruby version out of it if you haven't already

https://github.com/railwayapp/starters/blob/master/examples/rails-blog/Gemfile#L4

Milo123459: Ok, I plan to detect all frameworks by extracting Gems. I'll also add support for different Ruby versions.

PR #155, opened 2022-05-31
fallback to default go nix package on invalid version

jozsefsallai:
As requested in #150, I made it so that the Golang provider will fall back to the default "go" package when an invalid version is found in go.mod. Apologies for being late with this change, time zones are wack!

### DIFF
--- a/src/providers/go.rs
+++ b/src/providers/go.rs
@@ -5,7 +5,7 @@ use crate::nixpacks::{
     nix::Pkg,
     phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
 };
-use anyhow::{bail, Result};
+use anyhow::Result;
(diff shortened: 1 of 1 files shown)


src/providers/go.rs
```diff
@@ -110,7 +110,7 @@ fn version_number_to_pkg(version: &str) -> Result<Option<String>> {
 
     match matched_version {
         Some((_, pkg)) => Ok(Some(pkg.to_string())),
-        None => bail!("Go version {} is not available", version),
+        None => Ok(None),
```
Milo123459, line 113: Sorry if I'm missing something, but couldn't lots of the logic be cleaned up if this match statement always returned the version? Where it currently returns None should potentially return the default version.

PR #156, opened 2022-05-31
Improve Node logic

Milo123459:
This removes a lot of unneeded Results, replacing them mostly with Options. It also cleans up some of the code into it's own function for it to be used by other providers.

### DIFF
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -201,9 +201,9 @@ impl NodeProvider {
 
 fn version_number_to_pkg(version: &u32) -> String {
     if AVAILABLE_NODE_VERSIONS.contains(version) {
-        DEFAULT_NODE_PKG_NAME.to_string()
-    } else {
         format!("nodejs-{}_x", version)
+    } else {
+        DEFAULT_NODE_PKG_NAME.to_string()
     }
 }
 
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -142,7 +142,7 @@ impl NodeProvider {
         if let Some(node_pkg) = parse_regex_into_pkg(&re, node_version) {
             return Ok(Pkg::new(node_pkg.as_str()));
         }
-        println!("No package could be matched, defaulting to {DEFAULT_NODE_PKG_NAME}");
+        println!("No node version could be matched, defaulting to {DEFAULT_NODE_PKG_NAME}");
 
         Ok(Pkg::new(DEFAULT_NODE_PKG_NAME))
     }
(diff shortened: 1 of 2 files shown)


src/providers/node.rs
```diff
@@ -163,48 +133,89 @@ impl NodeProvider {
...
+        }
     }
 }
 
-fn version_number_to_pkg(version: &u32) -> Result<Option<String>> {
-    if AVAILABLE_NODE_VERSIONS.contains(version) {
-        Ok(Some(format!("nodejs-{}_x", version)))
+fn version_number_to_pkg(version: &u32) -> String {
+    if version % 2 == 1 {
+        String::from("nodejs")
```
JakeCooper, line 203: I think we should alert if we fallback from a non-stable version?

JakeCooper (approved):
Looks good one tiny comment!

src/providers/node.rs
```diff
@@ -163,48 +133,89 @@ impl NodeProvider {
...
```
wyzlle, line 203:
```suggestion
        DEFAULT_NODE_PKG_NAME.to_string()
```

JakeCooper: Yea good call

Milo123459: `DEFAULT_NODE_PKG_NAME` is just `nodejs`?

src/providers/node.rs
```diff
@@ -163,48 +134,89 @@ impl NodeProvider {
...
 }
 
-fn version_number_to_pkg(version: &u32) -> Result<Option<String>> {
+fn version_number_to_pkg(version: &u32) -> String {
     if AVAILABLE_NODE_VERSIONS.contains(version) {
-        Ok(Some(format!("nodejs-{}_x", version)))
+        format!("nodejs-{}_x", version)
     } else {
-        bail!("Node version {} is not available", version);
+        println!("No node version could be matched, defaulting to {DEFAULT_NODE_PKG_NAME}");
```
wyzlle, line 205:
```suggestion
        println!("\x1b[93mWARNING: No node version could be matched, defaulting to {DEFAULT_NODE_PKG_NAME}\x1b[0m");
```

src/providers/node.rs
```diff
@@ -7,12 +7,12 @@ use crate::nixpacks::{
     nix::Pkg,
     phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
 };
-use anyhow::{bail, Result};
+use anyhow::Result;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 
+const DEFAULT_NODE_PKG_NAME: &'static &str = &"nodejs";
```
wyzlle, line 14:
```suggestion
const DEFAULT_NODE_PKG_NAME: &'static &str = &"nodejs-16_x";
```

Milo123459: Not needed

wyzlle: I believe this is required. If nix updates default version it will break nixpack's reproducibility.

PR #159, opened 2022-05-31
Provider assets

aleksrutins:
This pull request adds static assets support for providers. It uses a new `static_assets` method in `Provider` that returns a `HashMap` with filenames as keys and file contents as values. It puts files into `/assets` in the Docker container, and you can get the path to an asset by calling `app.asset_path(name: &str)`.

See #158

### DIFF
--- a/src/nixpacks/mod.rs
+++ b/src/nixpacks/mod.rs
@@ -100,7 +100,9 @@ impl<'a> AppBuilder<'a> {
         let build_phase = self.get_build_phase().context("Generating build phase")?;
         let start_phase = self.get_start_phase().context("Generating start phase")?;
         let variables = self.get_variables().context("Getting plan variables")?;
-        let static_assets = self.get_static_assets().context("Getting provider assets")?;
+        let static_assets = self
+            .get_static_assets()
+            .context("Getting provider assets")?;
 
         let plan = BuildPlan {
             version: Some(NIX_PACKS_VERSION.to_string()),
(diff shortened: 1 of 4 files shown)


src/nixpacks/mod.rs
```diff
@@ -381,6 +396,19 @@ impl<'a> AppBuilder<'a> {
         File::create(dockerfile_path.clone()).context("Creating Dockerfile file")?;
         fs::write(dockerfile_path, dockerfile).context("Writing Dockerfile")?;
 
+        let static_assets_path = PathBuf::from(dest).join(PathBuf::from("assets"));
+        fs::create_dir_all(&static_assets_path).context("Creating static assets folder")?;
```
coffee-cup, line 400: If there are no assets I don't think the `assets/` directory should be created

src/providers/mod.rs
```diff
@@ -13,6 +15,8 @@ pub mod node;
 pub mod python;
 pub mod rust;
 
+pub type StaticAssets = HashMap<String, String>;
```
coffee-cup, line 18: nit: Move this type out the providers module. Maybe to `app.rs`?

coffee-cup (commented):
This is awesome! A few small comments.

Also, this doesn't seem to work with sub directories (i.e. specifying `dir/hello.txt` as the filename). It errors when attempting to copy the files. I think we should support this case or show a better error than

image

aleksrutins:
Should all hopefully be resolved now.

coffee-cup (approved):
Thanks :D

PR #161, opened 2022-06-02
Decouple plan generation from image creation

coffee-cup:
A fairly significant refactor that decouples the generation of a build plan from creating an image. Both are defined via traits. The image creation is setup such that if we wanted to use [something other than docker](https://github.com/railwayapp/nixpacks/discussions/121), we could.

The plan generator takes an app and environment and produces a build plan.

```rust
pub trait PlanGenerator {
    fn generate_plan(&mut self, app: &App, environment: &Environment) -> Result<BuildPlan>;
}
```

A builder takes a build plan and creates an image

```rust
pub trait Builder {
    fn create_image(&self, app_source: &str, plan: &BuildPlan) -> Result<()>;
}
```

The main library functions only care about these traits and the code is quite a bit cleaner.

I've also cleaned up the generate plan tests with a little helper.

_Note: I think this helps quite a bit but very open to suggestions on other improvements we could make_.


JakeCooper (approved):
Is there any "Why" beyond making it pluggable with another build system?

coffee-cup:
> Is there any "Why" beyond making it pluggable with another build system?

The previous `nixpacks/mod.rs` file was getting really large and messy as it was responsible for both detecting with the providers, generating the build plan, and creating an image. It was fairly hard to follow and even harder to update. Decoupling it makes the responsibilities of each part of nixpacks a lot more clear and manageable.

Issue #153, opened 2022-05-30
Django out-of-the-box support


JakeCooper:
Currently Django doesn't work out of the box because of the following error

```
Preparing metadata (setup.py): started
 
Preparing metadata (setup.py): finished with status 'error'
 
  error: subprocess-exited-with-error
 
× python setup.py egg_info did not run successfully.
│ exit code: 1
╰─> [23 lines of output]
running egg_info
creating /tmp/pip-pip-egg-info-vudn72c8/psycopg2.egg-info
writing /tmp/pip-pip-egg-info-vudn72c8/psycopg2.egg-info/PKG-INFO
writing dependency_links to /tmp/pip-pip-egg-info-vudn72c8/psycopg2.egg-info/dependency_links.txt
writing top-level names to /tmp/pip-pip-egg-info-vudn72c8/psycopg2.egg-info/top_level.txt
writing manifest file '/tmp/pip-pip-egg-info-vudn72c8/psycopg2.egg-info/SOURCES.txt'
 
Error: pg_config executable not found.
 
pg_config is required to build psycopg2 from source.  Please add the directory
containing pg_config to the $PATH or specify the full executable path with the
option:
 
python setup.py build_ext --pg-config /path/to/pg_config build ...
 
or with the pg_config option in 'setup.cfg'.
 
If you prefer to avoid building psycopg2 from source, please install the PyPI
'psycopg2-binary' package instead.
 
For further information please check the 'doc/src/install.rst' file (also at
<[https://www.psycopg.org/docs/install.html>](https://www.psycopg.org/docs/install.html%3E);).
 
[end of output]
 
note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
 
× Encountered error while generating package metadata.
╰─> See above for output.
 
note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
WARNING: You are using pip version 22.0.4; however, version 22.1.1 is available.
You should consider upgrading via the '/opt/venv/bin/python -m pip install --upgrade pip' command.
The command '/bin/bash -o pipefail -c python -m venv /opt/venv && . /opt/venv/bin/activate && pip install -r requirements.txt' returned a non-zero code: 1
 
Error: Docker build failed
```

I THINK this can be resolved by adding the nixpkg `postgresql`, but unsure

So, we need to check that and then add a provider for Django which, if it detects some Django-y things (manage.py in root? something else?) it'll grab that dep

wyzlle:
[...]

PR #162, opened 2022-06-02
Django Provider

JakeCooper:
This PR adds a few things:

* Django support, including installing postgres and GCC so that it works with PG
* Addition of container spawning for Postgres + network creation

In the future, it might be good to use a library like [Bollard](https://github.com/fussybeaver/bollard). I tried but it's kinda annoying to have to write an image puller, create container, and start container all separately (though I suppose you could just abstract it)


JakeCooper:
Fixed #153

ndneighbor (approved):
Looks good to ship.

src/providers/python.rs
```diff
@@ -136,6 +155,33 @@ enum EntryPoint {
 }
 
 impl PythonProvider {
+    fn is_django(app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("manage.py"))
```
coffee-cup, line 159: I think this check should be a bit more detailed? Should it also check if `Django` is in the requirements file?

tests/docker_run_tests.rs
```diff
@@ -107,80 +122,218 @@ fn simple_build(path: &str) -> String {
...
+const POSTGRES_IMAGE: &str = "postgres";
+
+struct Network {
+    name: String,
+}
+
+fn attach_container_to_network(network_name: String, container_name: String) {
+    let mut docker_cmd = Command::new("docker");
+
+    docker_cmd
```
coffee-cup, line 134: big ol nit: These docker commands don't need to be set to a variable first. Can just `Command::new("docker").arg("...").....`

tests/docker_run_tests.rs
```diff
@@ -34,55 +36,68 @@ fn remove_containers(container_id: &str) {
...
-    remove_containers(&container_id);
+    stop_and_remove_container(container_id);
+}
+
+fn stop_and_remove_container(name: String) {
+    stop_containers(&name);
+    remove_containers(&name);
 }
 
+struct Config {
```
coffee-cup, line 57: Nit: rename to `DockerRunConfig`

tests/docker_run_tests.rs
```diff
@@ -34,55 +36,68 @@ fn remove_containers(container_id: &str) {
...
+    environment_variables: Option<serde_json::Value>,
```
coffee-cup, line 58: `serde_json::Value` feels like a weird type for this since it has nothing to do with Serde. Can it just be something like `HashMap` ? We also have an `EnvironmentVariables` type we could use.

coffee-cup (approved):
Few nits but it looks good (pending tests).

It will be really nice to have the ability to easily spin up postgres/networks/other docker things when testing these images

tests/docker_run_tests.rs
JakeCooper: Borrow checker got angry :(

JakeCooper: jk

tests/docker_run_tests.rs
JakeCooper: It's not techincally a run config cause it's also passed to and from run

Open to other names though it's a bit generic

tests/docker_run_tests.rs
JakeCooper: Changed to a hashmap

PR #163, opened 2022-06-02
Add staticfile provider

### LINKED ISSUE #77: Staticfile/HTML provider

wyzlle:
Fixes #77

### DIFF
--- a/examples/staticfile/Staticfile.auth
+++ b/examples/staticfile/Staticfile.auth
@@ -1,0 +1,1 @@
+railway:$apr1$21i3ye0o$d1H1ePSg8mD2cO5O2A9dG.
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -12,7 +12,7 @@ pub mod haskell;
 pub mod node;
 pub mod python;
 pub mod rust;
-
+pub mod staticfile;
 pub trait Provider {
     fn name(&self) -> &str;
     fn detect(&self, app: &App, _env: &Environment) -> Result<bool>;
--- a/src/providers/staticfile.rs
+++ b/src/providers/staticfile.rs
@@ -1,0 +1,117 @@
+use super::Provider;
+use crate::nixpacks::{
+    app::{App, StaticAssets},
+    environment::Environment,
+    nix::pkg::Pkg,
+    phase::{BuildPhase, SetupPhase, StartPhase},
+};
+use anyhow::Result;
+use indoc::formatdoc;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct Staticfile {
+    pub root: Option<String>,
+    pub directory: Option<String>,
+    pub gzip: Option<String>,
+    pub status_code: Option<HashMap<u32, String>>,
+}
+
+pub struct StaticfileProvider {}
+
+impl Provider for StaticfileProvider {
+    fn name(&self) -> &str {
+        "staticfile"
+    }
+
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("Staticfile"))
+    }
+
+    fn setup(&self, _app: &App, _env: &Environment) -> Result<Option<SetupPhase>> {
+        let pkg = Pkg::new("nginx");
+        Ok(Some(SetupPhase::new(vec![pkg])))
+    }
+
+    fn static_assets(&self, app: &App, _env: &Environment) -> Result<Option<StaticAssets>> {
+        let mut assets = StaticAssets::new();
+
+        let mut mime_types = "".to_string();
+        if app.includes_file("mime.types") {
+            assets.insert("mime.types".to_string(), app.read_file("mime.types")?);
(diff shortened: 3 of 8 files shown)


src/providers/mod.rs
```diff
@@ -12,7 +12,7 @@ pub mod haskell;
 pub mod node;
 pub mod python;
 pub mod rust;
-
+pub mod staticfile;
```
Milo123459, line 15: There should be a space after this

src/providers/staticfile.rs
```diff
@@ -0,0 +1,117 @@
...
+    nix::pkg::Pkg,
+    phase::{BuildPhase, SetupPhase, StartPhase},
+};
+use anyhow::Result;
+use indoc::formatdoc;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct Staticfile {
```
Milo123459, line 14: I'm not sure, should this be StaticFile?

src/providers/staticfile.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(Some(BuildPhase::new(
+            "mkdir /etc/nginx/ /var/log/nginx/ /var/cache/nginx/".to_string(),
+        )))
+    }
+
+    fn start(&self, app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        Ok(Some(StartPhase::new(format!(
+            "nginx -c {conf_location}",
+            conf_location = app.asset_path("nginx.conf")
+        ))))
```
Milo123459, line 115: Perhaps this can log a URL it would be listening on? Also check if it is railway and log RAILWAY_STATIC_URL.

src/providers/staticfile.rs
wyzlle: `Staticfile` is fine IMO.

src/providers/staticfile.rs
wyzlle: Nixpacks is vendor/platform independent. Idk how I feel about adding a check from railway. @coffee-cup what do you think about this?

src/providers/staticfile.rs
wyzlle: `Staticfile` is being treated as a single word [here](https://docs.cloudfoundry.org/buildpacks/staticfile/) I guess (like `Procfile`).

devtomio:
couldn't this be renamed to `NginxProvider`, so we could also make caddy and other static file providers as well

Milo123459:
I agree, but why would you want to be able to choose what webserver you use when you can't configure it?

wyzlle:
Staticfile provider is different from Nginx provider. Nginx has more customization options whereas Staticfile provider is no brainer just adding an empty file named `Staticfile` will work. I don't really see a reason behind adding a caddy provider.

src/providers/staticfile.rs
```diff
@@ -0,0 +1,117 @@
...
+        if app.includes_file("Staticfile.auth") {
+            assets.insert(".htpasswd".to_string(), app.read_file("Staticfile.auth")?);
+            auth_basic = format!(
+                "auth_basic\t\"Password Required\";\nauth_basic_user_file\t{};",
+                app.asset_path(".htpasswd")
+            );
+        }
+
+        let staticfile: Staticfile = app.read_yaml("Staticfile").unwrap_or_default();
+        let port = std::env::var("PORT").unwrap_or_else(|_| "80".to_string());
```
coffee-cup, line 56: environment variable reading should come from the `Environment` struct, not from reading the variable directly.

src/providers/staticfile.rs
```diff
@@ -0,0 +1,117 @@
...
+        let root = staticfile.root.unwrap_or_else(|| "/app".to_string());
+        let gzip = staticfile.gzip.unwrap_or_else(|| "on".to_string());
+        let directory = staticfile.directory.unwrap_or_else(|| "on".to_string());
```
coffee-cup, line 59: I think this should be off by default

src/providers/staticfile.rs
coffee-cup: I agree with @wyzlle. Nixpacks shouldn't know anything about Railway or where the container will be running

src/providers/staticfile.rs
wyzlle: directory, gzip or both?

coffee-cup: Just directory

aleksrutins (changes requested):

examples/staticfile/Staticfile.auth
```diff
@@ -0,0 +1 @@
+railway:$apr1$21i3ye0o$d1H1ePSg8mD2cO5O2A9dG.
```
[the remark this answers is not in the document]
coffee-cup: Can you make it nixpacks,nixpacks. Just so the user name and password are the same. Or add a comment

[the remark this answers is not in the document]
wyzlle: Done!

coffee-cup (approved):
Looks good!

Issue #78, opened 2022-05-07
Php provider


wyzlle:

JakeCooper:
Adding a $100 bounty for this provider. Payable in crypto or USD.

aleksrutins:
I'd be up for this.

PR #165, opened 2022-06-02
Php provider

aleksrutins:
#78
This also adds a `static_asset_list!` macro, which automatically builds a `HashMap`.


src/providers/php.rs
```diff
@@ -0,0 +1,203 @@
...
+        Ok(Some(SetupPhase::new(pkgs)))
+    }
+
+    fn install(
+        &self,
+        app: &crate::nixpacks::app::App,
+        env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::phase::InstallPhase>> {
+        let mut cmd = String::new();
+        cmd.push_str("sudo apt-get update && sudo apt-get install -y nginx");
```
coffee-cup, line 65: Nginx is available on Nix and we should install it from there instead. https://search.nixos.org/packages?channel=unstable&show=nginx&from=0&size=50&sort=relevance&type=packages&query=nginx

aleksrutins: See #183 &mdash; in order to serve PHP, the Nginx configuration needs access to the system FastCGI configuration, which is installed with Nginx. In order to know where that is, I need to access it from within the Nix configuration.

coffee-cup: I see. Can you provide an example `environment.nix` file that would work so we can start getting that capability integrated into the Nix pkg struct.

aleksrutins: I figured out how to install Nginx through Nix without having access to `environment.nix`, with the `nix-store` command.

aleksrutins:
Writing Docker run tests could be a bit of a challenge, for two reasons:
1. Nginx doesn't give any output to the command line until you send it a request.
2. Php-fpm takes a few seconds to warm up and stop giving 502 errors, so the test would have to wait ~10s before actually checking the content.

src/providers/php.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn install(
+        &self,
+        app: &crate::nixpacks::app::App,
+        env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::phase::InstallPhase>> {
+        let mut cmd = String::new();
+        cmd.push_str("mkdir -p /var/log/nginx && mkdir -p /var/cache/nginx");
+        let nodejs = NodeProvider {};
```
coffee-cup, line 68: This is really interesting and a good forcing function to allow matching multiple providers at once. This is okay for now but I think allowing more than one provider would clean this up quite a bit.

src/providers/php.rs
```diff
@@ -0,0 +1,211 @@
...
+                    .as_str(),
+            );
+        }
+        Ok(Some(InstallPhase::new(cmd)))
+    }
+
+    fn build(
+        &self,
+        app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
```
coffee-cup, line 93: nit: These can be moved to imports so we don't have to reference by `crate::nixpacks` everywhere

src/providers/php.rs
```diff
@@ -0,0 +1,211 @@
...
+        Ok(None)
+    }
+
+    fn start(
+        &self,
+        app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::phase::StartPhase>> {
+        Ok(Some(StartPhase::new(format!(
+            "([ -e /app/storage ] && chmod -R ugo+w /app/storage); perl {} {} /nginx.conf && (php-fpm -y {} & nginx -c /nginx.conf)",
```
coffee-cup, line 109: What does the `/app/storage` do?

src/providers/php.rs
```diff
@@ -0,0 +1,211 @@
...
+        );
+        if app.includes_file("artisan") {
+            vars.insert("IS_LARAVEL".to_string(), "yes".to_string());
+        }
+        Ok(Some(vars))
+    }
+}
+
+impl PhpProvider {
+    fn get_php_package(&self, app: &crate::nixpacks::app::App) -> anyhow::Result<Option<String>> {
```
coffee-cup, line 147: The PHP versions are duplicated in a few places. I think there should be a single constant map of php version number to nix package and another constant for the default php version. It would make looking up the version and nix package cleaner.

src/providers/php.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn environment_variables(
+        &self,
+        app: &crate::nixpacks::app::App,
+        _env: &crate::nixpacks::environment::Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::environment::EnvironmentVariables>> {
+        let mut vars = EnvironmentVariables::new();
+        vars.insert(
+            "PHP_VERSION".to_string(),
```
coffee-cup, line 135: What is this variable needed for?

coffee-cup (commented):
I am seeing some issues when running both the vanilla PHP and laravel examples. Adding some docker run tests should catch this.

image

src/providers/php.rs
aleksrutins: `/app/storage` is where Laravel stores built templates, logs, and anything else it needs to store.

src/providers/php.rs
aleksrutins: I had been using this to find the FastCGI path, but I'm doing that differently now, so I'll remove it.

coffee-cup (approved):
Looks good pending lints!

PR #172, opened 2022-06-02
Improvements to staticfile provider

### LINKED ISSUE #170: Static file root should be relative to app source
When specifying a directory to serve with the static file provider, you have to prefix with `/app`. The user should not have to know anything about the directory structure inside the docker container. Specifying `root: public` should actually serve `/app/public`

wyzlle:
- fixes #170 
 - Ability to configure root using environment variable
 - Auto serve from `public`, `index` or `dist` if found
 - fix `includes_file` API
 - add `includes_directory` API

### DIFF
--- a/src/providers/staticfile.rs
+++ b/src/providers/staticfile.rs
@@ -53,7 +53,7 @@ impl Provider for StaticfileProvider {
         }
 
         let staticfile: Staticfile = app.read_yaml("Staticfile").unwrap_or_default();
-        let root = staticfile.root.unwrap_or_else(|| "/app".to_string());
+        let root = staticfile.root.unwrap_or_else(|| "".to_string());
         let gzip = staticfile.gzip.unwrap_or_else(|| "on".to_string());
         let directory = staticfile.directory.unwrap_or_else(|| "off".to_string());
         let status_code = staticfile.status_code.unwrap_or_default();
@@ -80,7 +80,7 @@ impl Provider for StaticfileProvider {
             server {{
                 listen    0.0.0.0:80;
                 gzip  	  {gzip};
-                root	  {root};
+                root	  /app/{root};
                 location / {{
                     {auth_basic}
                     autoindex {directory};
--- a/src/providers/staticfile.rs
+++ b/src/providers/staticfile.rs
@@ -128,9 +128,9 @@ impl Provider for StaticfileProvider {
 impl StaticfileProvider {
     pub fn get_root(app: &App, env: &Environment, staticfile_root: String) -> String {
         let mut root = "";
-        if env.get_variable("NIXPACKS_STATICFILE_ROOT").is_some() {
-            root = env.get_variable("NIXPACKS_STATICFILE_ROOT").unwrap();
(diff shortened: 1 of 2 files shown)


JakeCooper (approved):
Awesome looks great!

src/providers/staticfile.rs
```diff
@@ -116,3 +124,21 @@ impl Provider for StaticfileProvider {
         ))))
     }
 }
+
+impl StaticfileProvider {
+    pub fn get_root(app: &App, env: &Environment, staticfile_root: String) -> String {
+        let mut root = "";
+        if env.get_variable("NIXPACKS_STATICFILE_ROOT").is_some() {
+            root = env.get_variable("NIXPACKS_STATICFILE_ROOT").unwrap();
+        } else if staticfile_root.is_empty() {
```
coffee-cup, line 133: Should this check if the root is not empty?

wyzlle: Oops, My bad. Yeah, it should check if root is not empty.

devtomio (changes requested):
shouldn't we use [`std::path::Path::is_file`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.is_file) and [`std::path::Path::is_dir`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.is_dir) as those don't return `Result`s

src/providers/staticfile.rs
```diff
@@ -116,3 +124,21 @@ impl Provider for StaticfileProvider {
         ))))
     }
 }
+
+impl StaticfileProvider {
+    pub fn get_root(app: &App, env: &Environment, staticfile_root: String) -> String {
+        let mut root = "";
+        if env.get_variable("NIXPACKS_STATICFILE_ROOT").is_some() {
+            root = env.get_variable("NIXPACKS_STATICFILE_ROOT").unwrap();
```
devtomio, line 132:
```suggestion
        if let Some(staticfile_root) = env.get_variable("NIXPACKS_STATICFILE_ROOT") {
            root = staticfile_root;
```

devtomio (changes requested):

devtomio (approved):

coffee-cup (approved):

src/providers/staticfile.rs
```diff
@@ -116,3 +124,21 @@ impl Provider for StaticfileProvider {
         ))))
     }
 }
+
+impl StaticfileProvider {
+    pub fn get_root(app: &App, env: &Environment, staticfile_root: String) -> String {
+        let mut root = "";
+        if let Some(staticfile_root) = env.get_variable("NIXPACKS_STATICFILE_ROOT") {
```
coffee-cup, line 131: nit/for future: Can use `get_config_variable` that automatically prefixes with "NIXPACKS_"

wyzlle: sure, lemme make that change.

Milo123459: Changing how that function works should be in a different PR to update all the current uses of. NIXPACKS_ variables.

PR #176, opened 2022-06-03
Add swift provider

### LINKED ISSUE #151: Swift provider
Server-side Swift is a thing; see for example https://vapor.codes.

devtomio:
closes #151


.vscode/launch.json
```diff
@@ -0,0 +1,40 @@
+{
```
Milo123459, line 1: This file should be deleted.

src/providers/swift.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    fn build(&self, _app: &App, _env: &Environment) -> Result<Option<BuildPhase>> {
+        Ok(Some(BuildPhase::new(
+            "swift build -c release --static-swift-stdlib".to_owned(),
+        )))
+    }
+
+    fn start(&self, app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        let raw_paths = app.find_files("Sources/**/main.swift")?;
```
Milo123459, line 35: Is the directory Sources configurable?

src/providers/swift.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    fn environment_variables(
+        &self,
+        _app: &App,
+        _env: &Environment,
+    ) -> Result<Option<EnvironmentVariables>> {
+        let mut variables = EnvironmentVariables::default();
+
+        // https://stackoverflow.com/a/62271027
+        variables.insert("DEBIAN_FRONTEND".to_owned(), "noninteractive".to_owned());
```
Milo123459, line 67: Just a thought: perhaps to_string should be used for clarity?

Milo123459 (changes requested):

Milo123459:
This branch seems kind of messed up... there are commits merging all commits from master.

src/providers/swift.rs
devtomio: I think not?
https://forums.swift.org/t/customizing-location-of-source-directory/725/4

src/providers/swift.rs
devtomio: done

.vscode/launch.json
devtomio: ignored it instead in `,gitignore`, my vscode swift extension keeps regenerating the file

Milo123459:
Did you make this branch on master? 😅

Milo123459:
Do you write the code on master then make a new branch, or make a new branch then write the code?

devtomio:
> Did you write the code on master then make a new branch, or make a new branch then write the code?

the latter

Milo123459:
That's weird how that pull commit got in your branch then

src/providers/swift.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn start(&self, app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        let raw_paths = app.find_files("Sources/**/main.swift")?;
+        let paths = raw_paths
+            .iter()
+            .filter(|&path| !path.to_string_lossy().contains(".build"))
+            .collect::<Vec<_>>();
+
+        let path = match paths.first() {
+            Some(path) => path.to_string_lossy().to_string(),
+            None => return Ok(None),
```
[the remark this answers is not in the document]
Milo123459: Oh, I was getting confused with where this was. Reviewing on mobile is hard haha

src/providers/swift.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("Package.swift")
+            || (app.includes_file("Package.swift") && app.includes_file("Package.resolved")))
+    }
+
+    fn setup(&self, _app: &App, _env: &Environment) -> Result<Option<SetupPhase>> {
+        let pkgs = vec![
+            Pkg::new("clang_13"),
+            Pkg::new("python27Full"),
```
JakeCooper, line 26: Does Swift really need 2.7? smh

devtomio: switched to python 3.8

src/providers/swift.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::nixpacks::{
+    app::App,
+    environment::Environment,
+    nix::pkg::Pkg,
+    phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
+};
+use anyhow::Result;
+use indoc::formatdoc;
+
+static LATEST_SWIFT: &str = "5.6.1";
```
Milo123459, line 11: I believe this should be named DEFAULT_SWIFT_VERSION.

src/providers/swift.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        // https://forums.swift.org/t/which-clang-package-should-we-install/20542/14
+        let install_cmd = formatdoc! {"
+        sudo apt-get update && \
+        sudo apt-get install -y build-essential clang libsqlite3-0 libncurses6 libcurl4 libxml2 libatomic1 libedit2 libsqlite3-0 libcurl4 libxml2 libbsd0 libc6-dev && \
+        wget -q {download_url} && \
+        tar -xf {name}.tar.gz && \
+        sudo mv {name} /usr/share/swift
+        ",
+        name=name,
```
Milo123459, line 43: I'm surprised cargo fmt didn't pick this up, but could it be name = name? And same for the other ones.

src/providers/swift.rs
```diff
@@ -0,0 +1,147 @@
...
+        Ok(Some(BuildPhase::new(
+            "swift build -c release --static-swift-stdlib".to_string(),
+        )))
+    }
+
+    fn start(&self, app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        let raw_paths = app.find_files("Sources/**/main.swift")?;
+        let paths = raw_paths
+            .iter()
+            .filter(|&path| !path.to_string_lossy().contains(".build"))
```
Milo123459, line 64: Can't we just check if the path name is .build?

Milo123459 (changes requested):

src/providers/swift.rs
devtomio: Can't actually, e.g. you're in the `/workspaces/nixpacks` directory, `find_files` returns the full path of the files that match the glob and dependencies also have `main.swift` files e.g. `/workspaces/nixpacks/examples/swift-vapor/.build/checkouts/swift-collections/Benchmarks/benchmark-tool/main.swift`

src/providers/swift.rs
devtomio: done

src/providers/swift.rs
devtomio: done

src/providers/swift.rs
Milo123459: Use .has_match to check if the folder exists. I think this should work for directories.

devtomio: How? IIRC you can't use exclude patterns on glob and what's exactly wrong with iterating over the paths?

Milo123459: In the run code later down the line, you just do ./build so I thought that logic could be applied here.

devtomio: Oh, that logic cant be applied sadly, since the executable target is configured via the `Package.swift` file and I dont how to parse that.

Milo123459: Could a regex be used?

devtomio: > Could a regex be used?

Possibly, but I'm not really well-versed in regex.

devtomio: Also the field varies on different version so it'll be quite hard i think

Milo123459: That's fine. I can potentially give you a regex.

devtomio: If I may ask; what's wrong with my current method of finding the executable name? Since there can be only one `main.swift` file anyway.

Milo123459: From my knowledge, any file can be the mainfile. This is why I thought of making it dynamic so it would support more cases, if that makes sense.

devtomio: sure then up to you lol

devtomio: resolved in a discord thread

src/providers/swift.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    fn install(&self, app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
+        let version = SwiftProvider::get_swift_version(&app)?;
+        let (download_url, name) = SwiftProvider::make_download_url(version);
+
+        // https://forums.swift.org/t/which-clang-package-should-we-install/20542/14
+        let install_cmd = formatdoc! {"
+        sudo apt-get update && \
+        sudo apt-get install -y build-essential clang libsqlite3-0 libncurses6 libcurl4 libxml2 libatomic1 libedit2 libsqlite3-0 libcurl4 libxml2 libbsd0 libc6-dev && \
```
coffee-cup, line 38: I am very hesitant to add a provider that installs random packages from apt-get like this. What is wrong with the Swift from nixpkgs? It is available on unstable https://search.nixos.org/packages?channel=unstable&show=swift&from=0&size=50&sort=relevance&type=packages&query=swift

coffee-cup (changes requested):

src/providers/swift.rs
devtomio: It's broken! [Uploading u2{95c13ee} (3557aab1-84a7-4f02-b35b-aae311858d1d id=0x7f0a02b4) …]()

devtomio: old versions also don't work

coffee-cup: I think instead of installing Swift from source, which slightly goes against the philosophy of nixpacks, we should instead fix the upstream Swift nix expression or create an overlay (similar to the [Rust overlay](https://github.com/oxalica/rust-overlay)) that has working versions of Swift. Swift is not necessary to reach parity with the buildpacks and I would like to avoid making this a standard way of installing language dependencies. However, I do really appreciate this PR and the effort that was put into it.

src/providers/swift.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut setup_phase = SetupPhase::new(vec![Pkg::new("swift")]);
+        let mut variables = EnvironmentVariables::default();
+        let swift_version = SwiftProvider::get_swift_version(app)?;
+        let rev = version_number_to_rev(&swift_version)?;
+
+        if let Some(rev) = rev {
+            setup_phase.set_archive(rev.to_string());
+        }
+
+        variables.insert("NIXPKGS_ALLOW_BROKEN".to_string(), 1.to_string());
```
coffee-cup, line 50: This shouldn't be needed anymore now that this variable is set in the base image. Can you try rebasing against master and checking.

src/providers/swift.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    fn build(&self, app: &App, _env: &Environment) -> Result<Option<BuildPhase>> {
+        let name = SwiftProvider::get_executable_name(app)?;
+        let build_cmd = format!(
+            "\
+        CC=clang++ \
+        CPATH=~/.nix-profile/include \
+        LIBRARY_PATH=~/.nix-profile/lib \
+        QTDIR=~/.nix-profile \
```
wyzlle, line 79: Variables should be handled via `environment_variables` function. Example: [Node provider](https://github.com/railwayapp/nixpacks/blob/7d8f3d718f7c0e75e20b2a51e7496c6048d71d78/src/providers/node.rs#L67-L73)

wyzlle (changes requested):

src/providers/swift.rs
devtomio: This is just for the build & the `clang` linker tho. Setting them globally may cause possible unwanted behavior.

wyzlle: What sort of unwanted behaviors? Mind providing an example where this can be reproduced?

devtomio: Nevermind then. The `CPATH`, `LIBRARY_PATH`, and the `QTDIR` variables should be probably set in the base image.

ref: https://nixos.wiki/wiki/Nix_vs._Linux_Standard_Base#Build_and_install_from_Source

src/providers/swift.rs
```diff
@@ -0,0 +1,177 @@
+use super::Provider;
+use crate::nixpacks::{
+    app::App,
+    environment::Environment,
+    nix::pkg::Pkg,
+    phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
+};
+use anyhow::{bail, Result};
+
+static DEFAULT_SWIFT_VERSION: &str = "5.5.3";
```
Milo123459, line 10: It looks like this version isn't actually in `AVAILABLE_SWIFT_VERSIONS`...

src/providers/swift.rs
```diff
@@ -0,0 +1,177 @@
...
+        // Safe to unwrap, because the path was filtered by the glob expression
+        let pos = names.iter().position(|&n| n == "Sources").unwrap();
+
+        names.drain(0..pos);
+
+        Ok(names[1].to_string())
+    }
+}
+
+fn version_number_to_rev(version: &str) -> Result<Option<String>> {
```
Milo123459, line 153: This should be `impl`ed to `SwiftProvider`

Milo123459, line 153: This function shouldn't return a `Result`, as there is nowhere in the code where it can return an `Err`.

Milo123459 (changes requested):
Just a few nits

src/providers/swift.rs
devtomio: The current version on nixpkgs is 5.5.3, thus I see no need to add it there.

src/providers/swift.rs
devtomio: I was just following the pattern that other providers have. For example in the go provider.

Milo123459: Most providers have moved away from this, and if there are any that haven't, I'll be making a PR to make sure everything is styled like this.

Milo123459 (approved):
LGTM

src/providers/swift.rs
```diff
@@ -0,0 +1,175 @@
...
+pub struct SwiftProvider {}
+
+impl Provider for SwiftProvider {
+    fn name(&self) -> &str {
+        "swift"
+    }
+
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("Package.swift")
+            || (app.includes_file("Package.swift") && app.includes_file("Package.resolved")))
```
coffee-cup, line 39: The second half of this `||` is redundant since it will already have matched if `Package.swift` exists.

tests/docker_run_tests.rs
```diff
@@ -111,7 +110,7 @@ fn simple_build(path: &str) -> String {
         path,
         Vec::new(),
         &GeneratePlanOptions {
-            pin_pkgs: true,
+            pin_pkgs: false,
```
coffee-cup, line 113: We should be pinning the packages in the tests. Not pinning means that they could break if the latest nix unstable channel updates to a newer version of a package that we use (has happened before). If pinning doesn't work for Swift specifically then we shouldn't pin for just that test (use `create_docker_image` directly). Any idea why it doesn't work with pinning? It is not an uncommon thing to do and ideally all providers work.

devtomio: > Any idea why it doesn't work with pinning?

After further investigation, the swift versions `5.5.x` doesn't work at all.

src/providers/swift.rs
```diff
@@ -0,0 +1,174 @@
...
+            Pkg::new("zlib.dev"),
+        ]);
+
+        let swift_version = SwiftProvider::get_swift_version(app)?;
+        let rev = SwiftProvider::version_number_to_rev(&swift_version);
+
+        if let Some(rev) = rev {
+            setup_phase.set_archive(rev);
+        } else {
+            setup_phase.set_archive("c82b46413401efa740a0b994f52e9903a4f6dcd5".to_string());
```
coffee-cup, line 54: Can this be pulled from the `AVAILABLE_SWIFT_VERSIONS` above so it is not hardcoded

coffee-cup (commented):
Just one small comment but looks good otherwise!

src/providers/swift.rs
devtomio: done

Issue #183, opened 2022-06-05
Allow increased Nix package configuration in generated Nix expression


aleksrutins:
Currently, providers are not able to access the `environment.nix` generated by Nixpacks. However, this can cause some problems, because Nix uses weird hashed paths for things and the only way to get those paths (that I know of) is to use APIs that are available only to Nix expressions. For instance, the PHP provider needs FastCGI to be available in order to actually serve PHP. However, it can't find the socket, since it's at one of the aforementioned weird paths, instead of at `/var/run/php/php8.1-fpm.socket`, where the system NGINX configuration expects it to be.

JakeCooper:
Shouldn't we just be adding FastCGI as a dep and then letting nix do it's thing for de-duping/resolution?

coffee-cup:
Can you provide a more concrete example of how accessing `environment.nix` would help.

The providers are the things that generate the Nix expression so it is kind of a chicken and egg problem. If a dep is needed ideally it can be added (like Cooper mentioned) as a dep.

aleksrutins:
What I meant is specifically being able to write custom Nix code into `environment.nix`, not reading it. For instance, the [Nix documentation for `php-fpm`](https://nixos.wiki/wiki/Phpfpm) has a great big block of necessary configuration, separate from the packages.

I might be misunderstanding the purpose of `environment.nix`, though - is it for both packages and configuration, or just for packages?

coffee-cup:
Ah okay, that makes sense. Yes ideally we can provide additional config to Nix packages like this. The `Pkg` type currently has support for overrides and overlays. But in this case a lot more configuration is required. I think we can extend the `Pkg` type to support this so you can add a Nix package with predefined configuration to the build.

coffee-cup:
After looking at the php-fmp configuration a bit more it looks like it would only work for NixOS, which is not what nixpacks uses. I do still agree that additional configuration in the generated Nix expression is necessary.

coffee-cup:
Going to close this as there is nothing concrete that can be added. I am all for more configuration via Nix expressions, but a separate issue can be opened for each type of configuration requested.

PR #195, opened 2022-06-08
Canvas provider

wyzlle:
This PR adds required libraries and variables for [canvas](https://www.npmjs.com/package/canvas) projects.

### DIFF
--- a/examples/node-no-lockfile-canvas/index.js
+++ b/examples/node-no-lockfile-canvas/index.js
@@ -1,2 +1,2 @@
 const canvas = require("canvas")
-console.log("Hello from Node");
+console.log("Hello from Node canvas");
--- a/examples/node-no-lockfile-canvas/index.js
+++ b/examples/node-no-lockfile-canvas/index.js
@@ -1,0 +1,2 @@
+const canvas = require("canvas")
+console.log("Hello from Node");
--- a/examples/node-no-lockfile-canvas/index.js
+++ b/examples/node-no-lockfile-canvas/index.js
@@ -1,2 +1,2 @@
-const canvas = require("canvas")
+require("canvas");
 console.log("Hello from Node canvas");
--- a/examples/node-no-lockfile/index.js
+++ b/examples/node-no-lockfile/index.js
@@ -1,1 +1,0 @@
-console.log("Hello from Node");
--- a/examples/node-no-lockfile/index.js
+++ b/examples/node-no-lockfile/index.js
@@ -1,2 +1,1 @@
-const canvas = require('canvas');
 console.log("Hello from Node");
--- a/examples/node-no-lockfile/index.js
+++ b/examples/node-no-lockfile/index.js
@@ -1,1 +1,2 @@
+const canvas = require('canvas');
 console.log("Hello from Node");
--- a/src/nixpacks/images.rs
+++ b/src/nixpacks/images.rs
@@ -1,3 +1,3 @@
-pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654798583";
+pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654800960";
 
 pub static DEBIAN_SLIM_IMAGE: &str = "debian:bullseye-slim";
--- a/src/nixpacks/images.rs
+++ b/src/nixpacks/images.rs
@@ -1,3 +1,3 @@
-pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654618369";
+pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654798583";
 
 pub static DEBIAN_SLIM_IMAGE: &str = "debian:bullseye-slim";
--- a/src/nixpacks/plan/generator.rs
+++ b/src/nixpacks/plan/generator.rs
@@ -207,6 +207,7 @@ impl<'a> NixpacksBuildPlanGenerator<'a> {
                 img => Some(img.to_owned()),
             };
         }
+        
         Ok(start_phase)
     }
 
--- a/src/nixpacks/plan/generator.rs
+++ b/src/nixpacks/plan/generator.rs
@@ -207,7 +207,7 @@ impl<'a> NixpacksBuildPlanGenerator<'a> {
                 img => Some(img.to_owned()),
             };
         }
-        
+
         Ok(start_phase)
     }
 
--- a/src/nixpacks/plan/generator.rs
+++ b/src/nixpacks/plan/generator.rs
@@ -9,7 +9,7 @@ use crate::{
         phase::{BuildPhase, InstallPhase, SetupPhase, StartPhase},
         NIX_PACKS_VERSION,
     },
-    providers::{node::NodeProvider, Provider},
+    providers::Provider,
 };
 use anyhow::{bail, Context, Ok, Result};
 
@@ -206,9 +206,6 @@ impl<'a> NixpacksBuildPlanGenerator<'a> {
                 img if img.is_empty() => None,
                 img => Some(img.to_owned()),
             };
-        }
-        if NodeProvider::uses_canvas(app) {
-            start_phase.cmd = Some(format!("for pkg in $(ls /nix/store | grep \"util-linux-.*-lib$\"); do export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:\"/nix/store/$pkg/lib\"; done && {}", start_phase.cmd.unwrap_or_default()));
         }
         Ok(start_phase)
     }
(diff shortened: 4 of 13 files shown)


src/nixpacks/plan/generator.rs
```diff
@@ -207,7 +207,9 @@ impl<'a> NixpacksBuildPlanGenerator<'a> {
                 img => Some(img.to_owned()),
             };
         }
-
+        if NodeProvider::uses_canvas(app) {
+            start_phase.cmd = Some(format!("for pkg in $(ls /nix/store | grep \"util-linux-.*-lib$\"); do export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:\"/nix/store/$pkg/lib\"; done && {}", start_phase.cmd.unwrap_or_default()));
+        }
```
coffee-cup, line 212: The generator (or nixpacks in general) shouldn't know anything specific about the providers. This should be moved to the node provider start command.

coffee-cup (changes requested):

src/nixpacks/plan/generator.rs
wyzlle: Deployment will crash if someone has `Procfile` in their repo.

coffee-cup: Hmm then need to find a better way to handle Procfile start commands. Or handle util-linux natively in the base image. The node provider should not be imported and used like this. This doesn't seem like a bad thing to have in the base. Do you know how much larger it makes the image?

image

The providers currently have a way of specifying environment variables for the container. We could also expand that functionality so that they can add dynamic things like this?

wyzlle: As mentioned in #183 letting providers read/edit `environment.nix` directly would be better IMO. ([link](https://discourse.nixos.org/t/node2nix-issues/10762/2))

coffee-cup: The `environment.nix` file doesn't exist when the provider lifecycle methods are run so there is nothing to read/edit yet. I do think though that the providers need a way to inject various Nix snippets into the environment that gets run. Just need to land on an abstraction that will work in the majority of cases.

coffee-cup: Even just being able to edit `environment.nix` isn't enough. This is an example that would write the library path to a `env.sh` file under the default Nix profile

```nix
{ }:

let pkgs = import <nixpkgs> { };
in with pkgs;
let
  APPEND_LIBRARY_PATH = "${lib.makeLibraryPath [ libGL libuuid ]}";
  myProfile = writeText "env" ''
    export LD_LIBRARY_PATH="${APPEND_LIBRARY_PATH}:$LD_LIBRARY_PATH"
  '';

in buildEnv {
  name = "env";
  paths = [
    (runCommand "env" { } ''
      mkdir -p $out/etc/profile.d
      cp ${myProfile} $out/etc/profile.d/env.sh
    '')

    nodejs
  ];
}
```

But we would still need to somehow load that so those variables are available at runtime.

I think for the `LD_LIBRARY_PATH` variable specifically, we should have a `setup_phase.set_libraries` function that would set everything up automatically.

coffee-cup: I think this makes adding the libraries available much easier https://github.com/railwayapp/nixpacks/pull/197

examples/node-no-lockfile-canvas/index.js
```diff
@@ -0,0 +1,2 @@
+const canvas = require("canvas")
+console.log("Hello from Node");
```
coffee-cup, line 2:
```suggestion
console.log("Hello from Node canvas");
```

src/providers/node.rs
```diff
@@ -177,24 +182,34 @@ impl NodeProvider {
     pub fn get_nix_packages(app: &App, env: &Environment) -> Result<Vec<Pkg>> {
         let package_json: PackageJson = app.read_json("package.json")?;
         let node_pkg = NodeProvider::get_nix_node_pkg(&package_json, env)?;
-
+        let mut pkg = vec![node_pkg.clone()];
```
coffee-cup, line 185:
```suggestion
        let mut pkgs = vec![node_pkg.clone()];
```

src/providers/node.rs
```diff
@@ -177,24 +182,34 @@ impl NodeProvider {
...
         }
+        Ok(pkg)
+    }
+
+    pub fn uses_canvas(app: &App) -> bool {
+        let package_json = app.read_file("package.json").unwrap_or_default();
+        let lock_json = app.read_file("package-lock.json").unwrap_or_default();
+        let yarn_lock = app.read_file("yarn.lock").unwrap_or_default();
+        let pnpm_yaml = app.read_file("pnpm-lock.yaml").unwrap_or_default();
+        package_json.contains("canvas")
```
coffee-cup, line 209: I feel like this check is a bit too loose? Probably fine but it'll probably end up with quite a few false positives.

coffee-cup (approved):
Some minor nits but looks good

src/providers/node.rs
wyzlle: Improved it a bit. Now, nixpacks will check for `"canvas"` in `package.json` and `/canvas/` in lockfiles.

examples/node-no-lockfile-canvas/index.js
```diff
@@ -0,0 +1,2 @@
+const canvas = require("canvas")
```
devtomio, line 1:
```suggestion
require("canvas");
```

Setting it to a variable seems unnecessary as it isn't used.

devtomio (changes requested):

examples/node-no-lockfile/index.js
```diff
@@ -1 +0,0 @@
-console.log("Hello from Node");
```
Milo123459, line 1: Unrelated change?

src/nixpacks/images.rs
```diff
@@ -1,3 +1,3 @@
-pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654798583";
+pub static DEFAULT_BASE_IMAGE: &str = "ghcr.io/railwayapp/nixpacks:debian-1654800960";
```
Milo123459, line 1: This also seems unrelated. A rebase would remove this.

src/providers/node.rs
```diff
@@ -36,7 +36,12 @@ impl Provider for NodeProvider {
 
     fn setup(&self, app: &App, env: &Environment) -> Result<Option<SetupPhase>> {
         let packages = NodeProvider::get_nix_packages(app, env)?;
-        Ok(Some(SetupPhase::new(packages)))
+        let mut setup_phase = SetupPhase::new(packages);
+        if NodeProvider::uses_canvas(app) {
```
Milo123459, line 40: Idea: for the future (I'll implement) a method to check if a `package.json` has a dependency. Should check only `dependencies`.

tests/docker_run_tests.rs
```diff
@@ -238,9 +238,9 @@ fn test_node_custom_version() {
 
 #[test]
 fn test_node_no_lockfile() {
-    let name = simple_build("./examples/node-no-lockfile");
+    let name = simple_build("./examples/node-no-lockfile-canvas");
     let output = run_image(name, None);
-    assert!(output.contains("Hello from Node"));
+    assert!(output.contains("Hello from Node canvas"));
```
Milo123459, line 243:
```suggestion
    assert_eq!(output, "Hello from Node canvas");
```

examples/node-no-lockfile/index.js
wyzlle: Nope.

Milo123459: How is this not unrelated?

src/nixpacks/images.rs
wyzlle: How will a rebase remove this?

Milo123459: You're not changing the image in this PR, so a rebase would remove this change

examples/node-no-lockfile/index.js
wyzlle: We don't need this file anymore.

src/nixpacks/images.rs
wyzlle: I am updating the image to `debian-1654800960`(canvas requires that). Currently nixpacks uses `debian-1654798583` and that's not the latest release.

Milo123459: I thought I saw a PR get merged about updating this?

examples/node-no-lockfile/index.js
Milo123459: Alright. I have just seen that you have changed the example from no-lockfile to no-lockfile-canvas. I guess that's fine.

tests/docker_run_tests.rs
wyzlle: `npm run start` is used as start command here and that logs other stuff too.

src/nixpacks/images.rs
wyzlle: You saw that right but in that same PR jr made some changes to the base Dockerfile.

PR #202, opened 2022-06-10
Configure libraries using variables

wyzlle:
Libraries can now be configured via `NIXPACKS_LIBS` variable or the `--libs` argument.

### DIFF
--- a/src/nixpacks/phase.rs
+++ b/src/nixpacks/phase.rs
@@ -47,12 +47,11 @@ impl SetupPhase {
         self.archive = Some(archive);
     }
 
-    pub fn add_library(&mut self, lib: String) {
-        if let Some(mut libraries) = self.libraries.clone() {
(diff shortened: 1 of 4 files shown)


src/nixpacks/phase.rs
```diff
@@ -47,12 +47,11 @@ impl SetupPhase {
         self.archive = Some(archive);
     }
 
-    pub fn add_library(&mut self, lib: String) {
-        if let Some(mut libraries) = self.libraries.clone() {
-            libraries.push(lib);
-            self.libraries = Some(libraries);
+    pub fn add_library(&mut self, lib: Vec<String>) {
```
coffee-cup, line 50:
```suggestion
    pub fn add_libraries(&mut self, libs: Vec<String>) {
```

coffee-cup (approved):
nit to change function name but looks good besides that

PR #203, opened 2022-06-11
Ruby provider

### LINKED ISSUE #26: Ruby on Rails provider

wyzlle:
closes #26


src/providers/ruby.rs
```diff
@@ -0,0 +1,95 @@
...
+        Ok(Some(install_phase))
+    }
+
+    fn start(&self, _app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        Ok(None)
+    }
+}
+
+impl RubyProvider {
+    // fn detect_framework(&self, app: &App) -> String {
```
Milo123459, line 54: Just quick FYI, this should work by detecting gems in the gemfile.

Milo123459 (changes requested):

src/providers/ruby.rs
wyzlle: Not the best way to detect gems. Current logic is based on how buildpacks work. So, it should be fine.

Milo123459: What is a better way? That is what we decided on.

wyzlle: Current logic is better because buildpacks have used that for years now.

coffee-cup (approved):

src/nixpacks/phase.rs
```diff
@@ -12,6 +12,7 @@ pub struct SetupPhase {
     pub archive: Option<String>,
     pub libraries: Option<Vec<String>>,
     pub apt_pkgs: Option<Vec<String>>,
+    pub cmd: Option<Vec<String>>,
```
coffee-cup, line 15: Rename the variable to be more descriptive. Also, if this is a vector of commands, I think the commands for other phases should be a vector as well to be consistent.

```suggestion
    pub cmds: Option<Vec<String>>,
```

src/providers/ruby.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+impl RubyProvider {
+    fn get_start_command(&self, app: &App) -> String {
+        if app.includes_file("config/application.rb")
+            && app
+                .read_file("config/application.rb")
+                .unwrap_or_default()
+                .contains("Rails::Application")
```
coffee-cup, line 67: suggestion: Extract this check to a `is_rails_app` util

src/providers/ruby.rs
```diff
@@ -0,0 +1,163 @@
...
+            return Ok(format!("ruby-{}", value.get(1).unwrap().as_str()));
+        }
+        bail!("Please specify ruby's version in .ruby-version file")
+    }
+
+    fn get_bundler_version(&self, app: &App) -> String {
+        if app.includes_file("Gemfile.lock") {
+            let gemfile_lock = app.read_file("Gemfile.lock").unwrap_or_default();
+            let array_lock: Vec<&str> = gemfile_lock.split('\n').collect();
+            for line in 0..array_lock.len() {
```
coffee-cup, line 105: Can you add a descriptive comment here explaining what this is doing.

coffee-cup (approved):
Ruby provider looks good! And thanks for making the changes to all the other providers 😄

PR #209, opened 2022-06-13
Remove parsing of workspaces

### LINKED ISSUE #192: Support nohoist in yarn workspaces
https://classic.yarnpkg.com/blog/2018/02/15/nohoist/

If the `workspaces` field of `package.json` is shaped like the following

```ts
export type WorkspacesConfig = {
  packages?: Array<string>,
  nohoist?: Array<string>,
};
```

our current Node provider crashes. However, this is a completely valid package json and yarn workspace format that we should support.

Milo123459:
Removes the parsing of workspaces. Fixes #192.

I have decided to remove it because the parsed workspaces were actually doing nothing.

### DIFF
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -19,7 +19,6 @@ pub struct PackageJson {
     pub name: Option<String>,
     pub scripts: Option<HashMap<String, String>>,
     pub engines: Option<HashMap<String, String>>,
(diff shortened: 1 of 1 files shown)


coffee-cup (approved):
Thanks. It was used previously but can easily be removed now that it is no longer used.

PR #215, opened 2022-06-17
Install dev deps when using pnpm

Milo123459:
Currently, the pnpm provider didn't install dev dependencies. This PR changes it for consistency.


wyzlle (approved):

wyzlle:
We were wrong. Currently, `pnpm` installs `devDependencies` too. Couldn't find it in docs but while experimenting I figured out that pnpm auto-installs devDeps when `NPM_CONFIG_PRODUCTION` variable is set to false. Moreover, `node-pnpm` examples has typescript as a dev dependency and it works absolutely fine.

PR #226, opened 2022-06-25
Fix Python versions

nebulatgs:
The newly added Python versions only exist in unstable Nix
This PR adds support for stable Nix


JakeCooper:
We use unstable

nebulatgs:
> We use unstable

whoops, mb then. saw a user running into build issues and assumed this was why

PR #228, opened 2022-06-26
Verify if path is a file in app.find_files

### LINKED ISSUE #223: App.find_match() does not check for folders with file style names
[App.find_match()](https://github.com/railwayapp/nixpacks/blob/71ba2d256d454e3c01c2b42375ecb7426eab5829/src/nixpacks/app.rs#L82-L100) throws `Error: Is a directory (os error 21)` when folders have file style names.

Example dependency that causes this error: https://github.com/ethers-io/ethers.js/tree/master/packages/bytes

The package above has a folder named `src.ts`, causing [read_file](https://github.com/railwayapp/nixpacks/blob/71ba2d256d454e3c01c2b42375ecb7426eab5829/src/nixpacks/app.rs#L92) to throw the error with `**/*.ts` Regex pattern.

gustavo-dev:
Adds a check to verify if the path is a file.

Fixes #223

### DIFF
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -57,7 +57,7 @@ impl App {
             .into_iter()
             .filter_map(|result| result.ok()) // remove bad ones
             .map(|dir| dir.into_path()) // convert to paths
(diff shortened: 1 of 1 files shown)


JakeCooper (approved):
Thank you!

PR #229, opened 2022-06-26
use `std::env::consts::ARCH` instead of hardcoding

devtomio:


JakeCooper (approved):

tests/generate_plan_tests.rs
```diff
@@ -341,12 +342,11 @@ fn test_custom_rust_version() -> Result<()> {
...
 fn test_rust_rocket() -> Result<()> {
     let plan = simple_gen_plan("./examples/rust-rocket");
-    assert_eq!(
-        plan.build.unwrap().cmds,
-        Some(vec![
-            "cargo build --release --target x86_64-unknown-linux-musl".to_string()
-        ])
+    let cmd = format!(
+        "cargo build --release --target {}-unknown-linux-musl",
+        consts::ARCH
```
Milo123459, line 347: IMO, it would be better to explicitly import `ARCH` from the use that was added.

devtomio: done

PR #236, opened 2022-06-30
Support Deno Tasks/Fresh

JakeCooper:
Read the deno.json file and grab the start command from it if needed


src/providers/deno.rs
```diff
@@ -41,6 +53,18 @@ impl Provider for DenoProvider {
     }
 
     fn start(&self, app: &App, _env: &Environment) -> Result<Option<StartPhase>> {
+        // First check for a deno.json and see if we can rip the start command from there
+        if app.includes_file("deno.json") {
+            let deno_json: DenoJson = app.read_json("deno.json")?;
+
+            if let Some(tasks) = deno_json.tasks {
+                if let Some(start) = tasks.start {
```
coffee-cup, line 61: The Deno fresh start command uses `deno run` but to build we run [`deno compile`](https://deno.land/manual/tools/compiler) which "will compile the script into a self-contained executable". Does `deno run` use the compiled executable if it exists? Is `deno run` the recommended way of starting fresh in production?

JakeCooper: I wasn't sure, B U T, deno task start does seem to be the correct avenue for this

If there's a better way, we should update the examples. But this PR I think still stands cause we should be detecting the start command as `deno task start`, in the same way we detect `yarn start` etc

coffee-cup (approved):

src/providers/deno.rs
```diff
@@ -9,6 +9,18 @@ use crate::nixpacks::{
 };
 use anyhow::{Context, Result};
 use regex::Regex;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+
+pub struct DenoTasks {
+    pub start: Option<String>,
```
Milo123459, line 17: Not sure about this. I believe deno tasks can have other tasks aswell? (ie, db migrations)

src/providers/deno.rs
```diff
@@ -9,6 +9,18 @@ use crate::nixpacks::{
 };
 use anyhow::{Context, Result};
 use regex::Regex;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+
```
Milo123459, line 15: This blank line can be removed

Milo123459 (changes requested):

src/providers/deno.rs
JakeCooper: Yea that's fine. Serde should just drop it on the floor. Can add that stuff later if we wanna action on it

Milo123459: We should handle deno tasks like we handle package.json scripts IMO.

src/providers/deno.rs
```diff
@@ -9,6 +9,17 @@ use crate::nixpacks::{
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct DenoTasks {
+    pub start: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct DenoJson {
+    pub tasks: Option<DenoTasks>,
```
Milo123459, line 21:
```suggestion
    #[serde(skip_serializing_if="Option::None")]
    pub tasks: Option<DenoTasks>,
```

PR #248, opened 2022-07-01
Check for docker install

13r0ck:
Previously when docker wasn't installed nixpacks would close with
a generic error: `Error: No such file or directory (os error 2)`
This gives a more clear error message with a link how to install.


coffee-cup (approved):
Thanks!  🎉

PR #250, opened 2022-07-02
Fix missing command in django application

Mr-Destructive:
- Fixed a command by adding the `wsgi` module in the gunicorn command

### DIFF
--- a/docs/pages/docs/providers/python.md
+++ b/docs/pages/docs/providers/python.md
@@ -47,7 +47,7 @@ poetry install --no-dev --no-interactive --no-ansi
 if Django Application
 
 ```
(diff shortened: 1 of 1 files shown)


Mr-Destructive:
This is unrelated to the PR but I have an problem building the nixpacks for a django app.

I have a issue with running the nixpacks build command.
The docker build step is getting skipped and that's why the build fails, I tried adding the build-cmd parameter but it doesn't work

image

## With build-cmd
image

Does anyone have any suggestions or should I open an Issue for it?

JakeCooper (approved):
Thanks! Please open a separate issue for the other stuff with an example to reproduce!

PR #254, opened 2022-07-02
dont follow symlinks when copying over app

coffee-cup:


FarazPatankar (approved):

besteFinanzen:
Hi I just had the case where it would be great to have the option to follow symlinks. Is there a chance to have this as an option?

PR #266, opened 2022-07-07
Fix versions of npm, yarn and pnpm

JuanM04:
Fixed incorrect versions of npm, yarn and pnpm. The solution was creating a custom Nix Channel, mainly because there is a long-running issue inside nix about these packages not installing correctly:

- https://github.com/svanderburg/node2nix/issues/293
- https://github.com/NixOS/nixpkgs/issues/145432

The channel is currently hosted [here](https://github.com/JuanM04/nix-channel), but it'll be better if you host it


src/nixpacks/builder/docker.rs
```diff
@@ -317,6 +317,10 @@ impl DockerBuilder {
 
           WORKDIR {app_dir}
 
+          # Update channels
```
coffee-cup, line 320: I don’t think we should be referencing a third party channel for all builds, especially when the channel is outside of the main nixpacks repo. 

I also think the problem only existed for pnpm and npm/yarn worked as expected.

if possible I think we should fix this issue upstream vs installing node from a new channel.

coffee-cup (changes requested):

src/nixpacks/builder/docker.rs
coffee-cup: Alternatively we could use an overlay, which nixpacks already has support for. Similar to the rust provider https://github.com/oxalica/rust-overlay

JuanM04: > I don’t think we should be referencing a third party channel for all builds

I will be hosted by you. Furthermore, as you can see inside the channel, these new packages only downloads the package manager from NPM, unzips it and binds it.

I also thought of putting it inside the docker image, but it wouldn't be updated as often

coffee-cup (approved):
Works great! Thanks 😄

PR #277, opened 2022-07-08
Remove some deps

Milo123459:
This is the result of `cargo update && cargo upgrade`. It is good to keep dependencies up to date.


Milo123459:
Yeah. I've never used Renovate but I'm sure it could work. (Or just use dependabot)

coffee-cup (approved):

PR #278, opened 2022-07-08
Parallelize functions with `rayon`

Milo123459:
This adds a new dependency, `rayon`, that allows us to parallelize iteration over Vecs.

### DIFF
--- a/src/nixpacks/nix/pkg.rs
+++ b/src/nixpacks/nix/pkg.rs
@@ -1,4 +1,3 @@
-use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
@@ -23,7 +22,7 @@ impl Pkg {
         match &self.overrides {
             Some(overrides) => {
                 let override_string = overrides
-                    .par_iter()
+                    .iter()
                     .map(|(name, value)| format!("{} = {};", name, value))
                     .collect::<Vec<_>>()
                     .join(" ");
@@ -53,7 +52,7 @@ impl Pkg {
         match &self.overrides {
             Some(overrides) => {
                 let override_string = overrides
-                    .par_iter()
+                    .iter()
                     .map(|(name, value)| format!("{} = {}", name, value))
                     .collect::<Vec<_>>()
                     .join(", ");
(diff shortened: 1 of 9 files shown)


nebulatgs:
Does this provide a meaningful performance improvement?

Milo123459:
Yes, where things are parallel is often where there is a lot of data going in.

src/nixpacks/nix/pkg.rs
```diff
@@ -22,7 +23,7 @@ impl Pkg {
         match &self.overrides {
             Some(overrides) => {
                 let override_string = overrides
-                    .iter()
+                    .par_iter()
                     .map(|(name, value)| format!("{} = {};", name, value))
```
coffee-cup, line 27: I don't think we should be using rayon for small cases like this. `overrides` is at most going to be length 3 and all the map is doing is formating a string. I think the overhead of rayon will actually be slower.

coffee-cup (commented):
Can you notice any significant speed improvement from using rayon? I am not opposed to adding it but feels like most of the cases it is being used in this PR will have at most a vector of length 1-10. Not sure if it is worth adding the extra dependency for that.

src/nixpacks/nix/pkg.rs
Milo123459: Ah, thought overrides was going to be longer then that.

nebulatgs:
> Not sure if it is worth adding the extra dependency for that.

It is most definitely not worth it. Nixpacks is already faster than Docker can read a Dockerfile and produce a build context, so there is no reason to add another dep to (maybe) make it a few milliseconds faster.

Milo123459:
I cannot test locally, however, the largest effect will happen when a large number of environment variables are present. Parallel iteration will help massively.

Milo123459:
Rayon is already in the Cargo sourcetree, this isn't actually adding another dep.

coffee-cup:
> I cannot test locally, however, the largest effect will happen when a large number of environment variables are present. Parallel iteration will help massively.

I think we should test this before merging it in. All that is happening it those iter maps is formatting a string. The speed to map over and format 100 variables is negligible compared to building the image with Docker.



> Rayon is already in the Cargo sourcetree, this isn't actually adding another dep.

Yes but it doesn't appear in the final binary now.

PR #281, opened 2022-07-08
Don't search for package.json files in node_modules

coffee-cup:
If the directory being built has a node_modules directory, don't bother find package.json files in there

### DIFF
--- a/src/providers/node.rs
+++ b/src/providers/node.rs
@@ -263,7 +263,7 @@ impl NodeProvider {
 
         // Find package.json files with a "next build" build script and cache the associated .next/cache directory
         for file in package_json_files {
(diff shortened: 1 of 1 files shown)


src/providers/node.rs
```diff
@@ -263,10 +263,18 @@ impl NodeProvider {
 
         // Find package.json files with a "next build" build script and cache the associated .next/cache directory
         for file in package_json_files {
-            let json: PackageJson = app.read_json(file.to_str().unwrap())?;
+            // Don't fine package.json files that are in node_modules
```
wyzlle, line 266:
```suggestion
            // Don't find package.json files that are in node_modules
```

JakeCooper (approved):

PR #287, opened 2022-07-10
Replace `-` with `.` when specifying nix packages

Milo123459:
This replaces `-` with `.` when a package is passed to the `NIXPACKS_PKGS` environment variable. They have the same behaviour, but `-` would break the nix flake.

### DIFF
--- a/src/nixpacks/plan/generator.rs
+++ b/src/nixpacks/plan/generator.rs
@@ -111,7 +111,7 @@ impl<'a> NixpacksBuildPlanGenerator<'a> {
 
         let env_var_pkgs = environment
             .get_config_variable("PKGS")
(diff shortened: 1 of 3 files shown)


devtomio:
What do you mean by they have the same behavior? AFAIK, hyphens (`-`) aren't allowed in package names.

Milo123459:
They are and they aren't. Some old packages use it.

coffee-cup (approved):
Do you have some examples of where `-`'s need to be replaced by `.`?

Milo123459:
Not really, but I know it's happened before. This is also helpful for users, ie, if they are trying to add `openssl.dev` and use `openssl-dev`, it would break.

PR #288, opened 2022-07-10
feat: add supports for --platform

aarnphm:
Since nixpacks is using buildkit as backend, it is possible to pass in
`--platform` to set target platform for the generated docker container.

This would enable the target app to be built on different target platform.

I don't know what is the team's progress on supporting multiple architecture for target nixpacks.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -66,6 +66,13 @@ fn main() -> Result<()> {
                         .multiple_values(true),
                 )
                 .arg(
+                    Arg::new("platform")
+                        .long("platform")
+                        .help("Set target platform for your output image")
+                        .takes_value(true)
+                        .multiple_values(true),
+                )
+                .arg(
                     Arg::new("cache-key")
                         .long("cache-key")
                         .help(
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -65,6 +65,13 @@ fn main() -> Result<()> {
                         .takes_value(true)
                         .multiple_values(true),
                 )
+                .arg(
+                    Arg::new("platform")
+                        .long("platform")
+                        .help("Set target platform for your output image")
+                        .takes_value(true)
+                        .multiple_values(true),
```
Milo123459, line 73: Why can this take multiple values?

coffee-cup: Because you can build for multiple platforms at once

coffee-cup (approved):
> I don't know what is the team's progress on supporting multiple architecture for target nixpacks.

There is nothing currently planned but only because it hasn't been requested yet. This is a great start!

Can you please update the CLI reference docs with the new option. https://github.com/railwayapp/nixpacks/blob/main/docs/pages/docs/cli.md

src/main.rs
aarnphm: I believe for building multiple platform, we will also have to pass in `--push` and `--load`.

aarnphm:
@coffee-cup docs are updated

PR #290, opened 2022-07-11
Add support for bun

Milo123459:
This adds support for the Bun package manager.


JakeCooper:
I don't know if we should have first class support for bun until it actually has a production serving mode (or at least a `--no-watch` flag)

It's not gonna work well on the free plan for anything more than a hello world because of inode stuff

Milo123459:
> I don't know if we should have first class support for bun until it actually has a production serving mode (or at least a `--no-watch` flag)
> 
> It's not gonna work well on the free plan for anything more than a hello world because of inode stuff

It does have a production serving mode. HMR only happens in dev mode.

coffee-cup:
Can you please add a new example (or update an existing one) of a Bun server. Just so we can test that it works there.

Milo123459:
A bun webserver? I'm not really sure how we would test that.

aarnphm:
I thought bun is a compiler and transpiler?

coffee-cup (commented):
The `node-bun-web-server/node_modules` directory should be ignored

src/providers/node.rs
```diff
@@ -280,10 +280,12 @@ impl NodeProvider {
         let lock_json = app.read_file("package-lock.json").unwrap_or_default();
         let yarn_lock = app.read_file("yarn.lock").unwrap_or_default();
         let pnpm_yaml = app.read_file("pnpm-lock.yaml").unwrap_or_default();
+        let bun_lock = app.read_file("bun.lockb").unwrap_or_default();
```
coffee-cup, line 283: The bun lockfile is in binary format so I don't think this will work

Milo123459: Let me add another test to see.

Milo123459: I would test it but it seems `bun` just gets stuck on installing canvas.

Milo123459:
> I thought bun is a compiler and transpiler?

And a runtime :)

src/providers/node.rs
Milo123459: So this works for checking if the package is in the lockfile, it just seems Bun doesn't support NAPI (so canvas cannot be used).

tests/docker_run_tests.rs
```diff
@@ -270,6 +270,27 @@ fn test_pnpm() {
...
+    let output = run_image(name, None);
+    assert!(output.contains("Hello from Bun"));
+}
+
+#[test]
+fn test_bun_no_start() {
+    let name = simple_build("./examples/node-bun-no-start");
+    let output = run_image(name, None);
+    assert!(output.contains("Hello from Bun"));
+}
```
coffee-cup, line 285: I think we only need one of these docker tests since the generate plan test will cover the other case

src/providers/node.rs
coffee-cup: Let's remove this check then if canvas can't be used

coffee-cup (approved):

src/providers/node.rs
Milo123459: I think we should keep it for future proofing. It doesn't work because of a bug in bun.

Milo123459:
Suggestions applied. Anything else need to happen?

PR #299, opened 2022-07-12
Prettier plan output

coffee-cup:
Output the build plan inside of a box in a table format. This makes it a bit easier to visually parse and see what is going on.

image

### DIFF
--- a/src/nixpacks/plan/mod.rs
+++ b/src/nixpacks/plan/mod.rs
@@ -37,33 +37,39 @@ impl BuildPlan {
 
         let top_box = format!(
             "{}{} {} {}{}",
-            box_drawing::double::DOWN_RIGHT.dimmed(),
+            box_drawing::double::DOWN_RIGHT.cyan().dimmed(),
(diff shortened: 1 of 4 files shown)


src/nixpacks/plan/mod.rs
```diff
@@ -27,54 +32,127 @@ pub trait PlanGenerator {
...
+            second_column_width,
+            console::Alignment::Left,
+            None
+        )
+        .white(),
+        box_drawing::double::VERTICAL.dimmed()
+    );
+
+    for line in list_lines.iter().skip(1) {
+        output.push_str(&format!(
```
JakeCooper, line 141: nit: GitHub warning

JakeCooper (approved):

coffee-cup:
> How does this look when using railway on mobile? Surely the box looks bad and lop-sided?

With no wording wrapping it will work fine. Might have to make some adjustments after for word wrapping enabled but that is basically impossible to account for

Issue #237, opened 2022-06-30
Using multiple providers


ipmb:
I have a project that uses both Node.js and Python. With buildpacks, I could specify `heroku/nodejs` and `heroku/python` and it would do the right thing (Node build/install followed by Python build/install).

With nixpacks, the project is identified as Node.js automatically and does the right thing there, but if I add `--pkgs=python39` to `nixpack build` it only installs Python, but doesn't use the default build/install command.

Is there a way to "stack" providers similar to how it is done with buildpacks? If not, that would be nice to have so I don't have to manually specify the additional build/install commands.

JakeCooper:
> Is there a way to "stack" providers similar to how it is done with buildpacks? If not, that would be nice to have so I don't have to manually specify the additional build/install commands.

ATM no, you'd have to manually define those additional commands which I could see as being not so great. Simply haven't run into in a way that doesn't "rollup" to framework specific handling (e.g Ruby runs some node stuff but that's invoked from the Ruby provider)

JakeCooper:
I don't want to jinx it, but we might be able to accomplish this by simply:
- Changing `[matched_provider](https://github.com/railwayapp/nixpacks/blob/59ba5914e411b778b015da4bb6374826ffc16825/src/nixpacks/plan/generator.rs#L89)` from an Option to a Vec
- Writing a function to "join" each stage

Would be open to a PR or we can look at it once there's more demand in that category

ipmb:
This is a pretty common use case for Django (and I assume Python in general). Use a standard Node.js pipeline for your static frontend assets and then build your Python project. For Django, you'd typically run `manage.py collectstatic --noinput` at the end to grab all the Node-built files and put them where Django expects. The Heroku buildpack runs this collectstatic step by default, but it has a toggle to turn it off because it can be problematic in some scenarios.

Finkregh:
I think that's a common usecase for many web related projects.
One half the actual software and the other half some nodejs to build assets.

JakeCooper:
> This is a pretty common use case for Django (and I assume Python in general). Use a standard Node.js pipeline for your static frontend assets and then build your Python project. For Django, you'd typically run manage.py collectstatic --noinput at the end to grab all the Node-built files and put them where Django expects. The Heroku buildpack runs this collectstatic step by default, but it has a toggle to turn it off because it can be problematic in some scenarios.

Right. Like I said, the Ruby/Rails Provider generally handles this, so I wouldn't be opposed to adding one of:
[...]

PR #300, opened 2022-07-12
[PoC] New build plan format and generic phases

coffee-cup:
This PR implements a new build plan format that
- Enables providers to define more phase types besides setup, install, build, and start
- Simplifies the generation of Dockerfile snippets and writing of supporting app files (nix expressions)
- Paves the way so we can better support matching against multiple providers (#237)

_Note: a lot of this PR is still a WIP and it is not 1:1 feature complete with the previous version. Some things are still a bit messy. I am opening this up early for feedback and discussion on the higher-level ideas. The "New" prefix is also just for simplifying the implementation of this PR and will be removed in the final version._

# Generic and consistent phases

Previously the build plan was comprised of four phases
- Setup (install packages)
- Install (install language deps)
- Build (build the app)
- Start (run the app when the container starts)

These phases were set in stone and hard coded. The implementation for each phase was also completely different, but very similar to each other. This made updating phases a tedious and confusing task.

This PR implements two new phase types: `Phase` and `StartPhase`. 

The generic `Phase` type can be used for any step that occurs during the creation of the image during the build. This would encapsulate the current setup, install, and build phases. Going from the phase to a dockerfile or nix expression is the same for each setup/build/install without needing a separate custom implementation for each.

The phase can also "depend on" another phase. When creating the image we will topologically sort these phases and execute them in the correct order.

```rust
#[serde_with::skip_serializing_none]
#[derive(Serialize, Deserialize, Default, Clone, Debug)]
pub struct NewPhase {
    pub name: String,

    #[serde(rename = "dependsOn")]
    pub depends_on: Option<Vec<String>>,

    #[serde(rename = "nixPackages")]
    pub nix_pkgs: Option<Vec<Pkg>>,

    #[serde(rename = "nixLibraries")]
    pub nix_libraries: Option<Vec<String>>,

    #[serde(rename = "nixpacksArchive")]
    pub nixpacks_archive: Option<String>,

    #[serde(rename = "aptPackages")]
    pub apt_pkgs: Option<Vec<String>>,

    #[serde(rename = "commands")]
    pub cmds: Option<Vec<String>>,

    #[serde(rename = "onlyIncludeFiles")]
    pub only_include_files: Option<Vec<String>>,

    #[serde(rename = "cacheDirectories")]
    pub cache_directories: Option<Vec<String>>,
}
```

The `StartPhase` is slightly different since this specifies how the container will start.

```rust
#[serde_with::skip_serializing_none]
#[derive(Serialize, Deserialize, Default, Clone, Debug)]
pub struct NewStartPhase {
    pub cmd: Option<String>,

    #[serde(rename = "runImage")]
    pub run_image: Option<String>,

    #[serde(rename = "onlyIncludeFiles")]
    pub only_include_files: Option<Vec<String>>,
}
```

# Provider Implementation

With these new phases comes a new way to create providers. Instead of Nixpacks calling into each provider to get each phase, the provider returns a full build plan. For example, the deno provider.

```rust
fn get_build_plan(&self, app: &App, _env: &Environment) -> Result<Option<NewBuildPlan>> {
    let mut plan = NewBuildPlan::default();

    let mut setup_phase = NewPhase::new("setup");
    setup_phase.add_nix_pkgs(vec![Pkg::new("deno")]);
    plan.add_phase(setup_phase);

    if let Some(start_file) = DenoProvider::get_start_file(app)? {
        let mut build_phase = NewPhase::new("build");
        build_phase.depends_on_phase("setup");

        build_phase.add_cmd(format!(
            "deno cache {}",
            start_file
                .to_str()
                .context("Failed to convert start_file to string")?
        ));

        plan.add_phase(build_phase);
    };

    if let Some(start_cmd) = DenoProvider::get_start_cmd(app)? {
        let start_phase = NewStartPhase::new(start_cmd);
        plan.add_start_phase(start_phase);
    }

    let build_plan = NewBuildPlan::default();
    Ok(Some(build_plan))
}
```

For more complicated providers this can obviously be cleaned up and split out to `get_build_phase` functions.

I am also thinking of a few ways to improve ergonomics with macros but that can come after the plan format is solidified.

_Note: we can keep the current implementation of the providers for now and just apply a conversion from old format to new format._

# Output generation

As part of the image creation process, a few files need to be generated and saved to disk. This include things such as static assets, nix expressions, and the Dockerfile.

Previously that was all shoved into a single file and was hard to parse. In this PR each type is responsible for generating code for the Dockerfile and writing supporting files. The result is much cleaner image creation

```rust
let dockerfile = plan
    .generate_dockerfile(&self.options, env)
    .context("Generating Dockerfile for plan")?;

self.write_app(app_src, dest).context("Writing app")?;
self.write_dockerfile(dockerfile, dest)
    .context("Writing Dockerfile")?;
plan.write_supporting_files(&self.options, env, dest)
    .context("Writing supporting files")?;

let mut docker_build_cmd = self.get_docker_build_cmd(plan, name.as_str(), dest)?;

// Execute docker build
let build_result = docker_build_cmd.spawn()?.wait().context("Building image")?;
if !build_result.success() {
  bail!("Docker build failed")
}
```

# New build plan format

And finally, the new build plan format. I think this format will lend itself well to using multiple providers to build because the plans can just be merged together to make a new build plan. It was also possible with the previous format, but this is much more flexible.

```rust
#[serde_with::skip_serializing_none]
#[derive(Debug, Serialize, Deserialize, Default, Clone)]
pub struct NewBuildPlan {
    #[serde(rename = "nixpacksVersion")]
    pub nixpacks_version: Option<String>,

    pub variables: Option<EnvironmentVariables>,

    #[serde(rename = "staticAssets")]
    pub static_assets: Option<StaticAssets>,

    pub phases: Vec<NewPhase>,

    #[serde(rename = "startPhase")]
    pub start_phase: Option<NewStartPhase>,
}
```


aarnphm:
Is it possible to define system packages per architecture in this new build plan proposal?

coffee-cup:
> Is it possible to define system packages per architecture in this new build plan proposal?

What do you mean? Like something that the providers can do?

That is a separate issue/feature and can be added after with this build plan.

src/nixpacks/builder/docker/dockerfile_generation.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        let dockerfile = formatdoc! {"
+            FROM {base_image}
+            WORKDIR {APP_DIR}
+
+            {dockerfile_phases_str}
+
+            {start_phase_str}
+        "};
+
```
AhmedMozaly, line 102: Could this support docker multi-stage build? (one stage for creating artifact & other stage for running the artifact)

PR #301, opened 2022-07-12
fix(providers): installing haskell on arm container

### LINKED ISSUE #104: Haskell provider doesn't work on M1 macs
Running Nixpackgs on the `examples/haskell-stack` directory fails with error

```
Unable to find installation URLs for OS key: linux-aarch64-tinfo6
```

It seems that the Haskell toolchain cannot find binaries for the aarch64 architecture on Linux

aarnphm:
https://github.com/commercialhaskell/stack/issues/5709#issuecomment-1097459982 kudos here.

Fixes #104 

cc @coffee-cup @Milo123459

### DIFF
--- a/src/providers/haskell.rs
+++ b/src/providers/haskell.rs
@@ -11,6 +11,7 @@ use crate::nixpacks::{
 use anyhow::Result;
 
 use super::Provider;
+use std::env::consts::ARCH;
 
 pub struct HaskellStackProvider {}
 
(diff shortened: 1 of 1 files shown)


src/providers/haskell.rs
```diff
@@ -27,6 +27,16 @@ impl Provider for HaskellStackProvider {
...
+            "libnuma1".to_string(),
+            "libnuma-dev".to_string(),
+            "libtinfo-dev".to_string(),
+            "libtinfo5".to_string(),
+            "libc6-dev".to_string(),
+            "libtinfo6".to_string(),
+            "llvm-11".to_string(),
+            "clang".to_string(),
+            "ninja-build".to_string(),
+            "zlib1g-dev".to_string(),
```
coffee-cup, line 39: can these be installed only if arm is detected? Since it currently works for non-arm

aarnphm: sounds good I will update it accordingly.

coffee-cup (approved):
Thanks!

PR #311, opened 2022-07-14
Redirect nixpacks.com/install.sh to install script

coffee-cup:
- Redirect nixpacks.com/install.sh to install script
- Update installation instructions accordingly


JakeCooper (approved):

devtomio:
Why not just serve the script instead of redirecting?

coffee-cup:
> Why not just serve the script instead of redirecting?

So that it can still be listed at the root of the repo

PR #319, opened 2022-07-15
Fix python numpy and pandas

coffee-cup:
## What does this PR address?
- Add the `zlib` and `stdenv.cc.cc.lib` Nix packages as libraries if numpy is detected
- Docker test for build and running python projects that use numpy and pandas


JakeCooper (approved):

src/providers/python.rs
```diff
@@ -47,7 +47,15 @@ impl Provider for PythonProvider {
             pkgs.append(&mut vec![Pkg::new("postgresql"), Pkg::new("gcc")]);
         }
 
-        Ok(Some(SetupPhase::new(pkgs)))
+        let mut setup_phase = SetupPhase::new(pkgs);
+
+        // Numpy needs some C headers to be available
+        // stdenv.cc.cc.lib -> https://discourse.nixos.org/t/nixos-with-poetry-installed-pandas-libstdc-so-6-cannot-open-shared-object-file/8442/3
+        if PythonProvider::uses_numpy(app)? {
+            setup_phase.add_libraries(vec!["zlib".to_string(), "stdenv.cc.cc.lib".to_string()]);
```
Milo123459, line 55: Shouldn't this also add this to the nixpkgs to be installed?

src/providers/python.rs
```diff
@@ -305,6 +313,14 @@ impl PythonProvider {
                 .ok_or_else(|| anyhow::anyhow!("failed to load pyproject.toml"))?),
         ))
     }
+
+    fn uses_numpy(app: &App) -> Result<bool> {
+        Ok(app.includes_file("requirements.txt")
+            && app
+                .read_file("requirements.txt")?
+                .to_lowercase()
+                .contains("numpy"))
```
Milo123459, line 322: This should probably support things like poetry.lock

Milo123459 (changes requested):

src/providers/python.rs
coffee-cup: ya good point?

src/providers/python.rs
coffee-cup: No it is not necessary

Milo123459: Then what does add_libraries do? What is the difference?

src/providers/python.rs
```diff
@@ -305,6 +313,22 @@ impl PythonProvider {
...
+    fn uses_numpy(app: &App) -> Result<bool> {
+        let requirements_numpy = app.includes_file("requirements.txt")
+            && app
+                .read_file("requirements.txt")?
+                .to_lowercase()
+                .contains("numpy");
+
+        let project_numpy = app.includes_file("pyproject.toml")
+            && app
+                .read_file("pyproject.toml")?
```
Milo123459, line 326: @coffee-cup This should have also checked poetry.lock

PR #324, opened 2022-07-17
Add ability to read .env file

AhmedMozaly:
PR should add ability to read .env file if found in the directory


Cargo.toml
```diff
@@ -41,3 +41,4 @@ base64 = "0.13.0"
 console = "0.15.0"
 box_drawing = "0.1.2"
 textwrap = "0.15.0"
+dotenv-parser = "0.1.3"
```
Milo123459, line 44: Couldn't we just use the dotenv crate? Maybe I am missing something.

AhmedMozaly: Actually I was planing & this was pushed by mistake :)
Will close now

PR #325, opened 2022-07-18
Fix `clippy::pedantic`

devtomio:
## What does this PR address?

Fixes [`clippy::pedantic`](https://rust-lang.github.io/rust-clippy/master/)

## Before submitting:


src/nixpacks/app.rs
```diff
@@ -33,6 +33,7 @@ impl App {
     }
 
     /// Check if a file exists
+    #[must_use]
```
coffee-cup, line 36: What is the reasoning for having `must_use` here?

coffee-cup (commented):
I'm happy with most of these changes besides all the `must_use`'s. What is the reasoning for them? Would love if we didn't have to add them everywhere

devtomio:
> I'm happy with most of these changes besides all the `must_use`'s. What is the reasoning for them? Would love if we didn't have to add them everywhere

Seems it has many false positives, I'll remove them.

coffee-cup:
> must_use just means you must use the return value. Is that a bad thing?

It is not a bad thing I'd prefer a global rule for this instead of specifically adding `must_use` to a select set of functions. New functions are also likely to not get this attribute as outside contributors will be unaware that they need to add it.

coffee-cup (approved):

coffee-cup:
Can you please fix the conflict in `node.rs`. Besdies that it looks good to merge.

PR #336, opened 2022-07-20
Ruby assets precompile

coffee-cup:
## What does this PR address?

Run `bundle exec rake assets:precompile` if Rails app detected


docs/pages/docs/providers/ruby.md
```diff
@@ -20,7 +20,11 @@ If a `package.json` file is found then the dependencies are installed with the r
 
 ## Build
 
-_None_
+If a Rails application
```
Milo123459, line 23: If a rails application is detected should be how this is written

PR #339, opened 2022-07-20
Install binutils in base docker image

Milo123459:
## What does this PR address?

`ar` has been shown to not be installed in some cases and causes issues. This should fix it.

## Before submitting:

- [ ] Did you write tests to cover your changes? Did it pass locally when running `cargo test`?
- [ ] Have you run `cargo fmt`, `cargo check` and `cargo clippy` locally to ensure that CI passes?

### DIFF
--- a/base/debian/Dockerfile
+++ b/base/debian/Dockerfile
@@ -3,7 +3,7 @@ FROM debian:bullseye-20220622-slim
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get -y upgrade \
-  && apt-get install --no-install-recommends -y sudo locales curl xz-utils ca-certificates openssl make git pkg-config \
+  && apt-get install --no-install-recommends -y sudo locales curl xz-utils ca-certificates openssl make git pkg-config binutils \
   && apt-get clean && rm -rf /var/lib/apt/lists/* \
   && mkdir -m 0755 /nix && mkdir -m 0755 /etc/nix && groupadd -r nixbld && chown root /nix \
   && printf 'sandbox = false \nfilter-syscalls = false' > /etc/nix/nix.conf \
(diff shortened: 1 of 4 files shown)


coffee-cup:
What is the base image size increase with this? The last I checked it was fairly large. 

I think we should prefer to install this only as needed (for specific packages or providers) as opposed to first installing it in the base image. Does this even fix the Rust issues?

Milo123459:
> What is the base image size increase with this? The last I checked it was fairly large.
> 
> I think we should prefer to install this only as needed (for specific packages or providers) as opposed to first installing it in the base image. Does this even fix the Rust issues?

I'm not exactly sure how to check the increase.

Though, installing only when needed is extremely hard. We can't just read build scripts and know when a binutils command is being invocated. And, issues with `ar` not being installed would be fixed by this.

coffee-cup:
> I'm not exactly sure how to check the increase.

Build the base image locally before and after the change. I think it was like a 200MB increase.

> Though, installing only when needed is extremely hard. We can't just read build scripts and know when a binutils command is being invocated. And, issues with ar not being installed would be fixed by this.

That is fair. But this has never been an issue with Node, Python, or Ruby projects. So making all of those images 200MB larger isn't ideal. We could just install `binutils` for Rust projects.

Milo123459:
Before: 420MB
After: 432MB

I think that change is fine.

coffee-cup:
I still think we should add it to the Rust provider first before adding it to the base image. I am not opposed to add it to the base in the future. But the default shouldn't be "add to base image". That is how we end up with a massive base that just has everything.

Secondly, does this fix the Rust issues?

Milo123459:
I'll make this only apply to the rust provider now.

And yes, this fixes the rust issues.

Milo123459:
Alright, CI is passing and now only binutils is installed in the Rust provider.

coffee-cup (approved):

PR #342, opened 2022-07-21
Add Clojure provider

AhmedMozaly:
Adds Clojure support to Nixpacks, work still in progress but feedback is more than welcomed!


src/providers/clojure.rs
```diff
@@ -0,0 +1,38 @@
...
+    environment::Environment,
+    nix::pkg::Pkg,
+    phase::{BuildPhase, SetupPhase, StartPhase},
+};
+use anyhow::Result;
+pub struct ClojureProvider {}
+
+impl Provider for ClojureProvider {
+    fn name(&self) -> &str {
+        "Clojure"
```
Milo123459, line 13: nit: should be lowercase

src/providers/clojure.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("project.clj"))
+    }
+
+    fn setup(&self, _app: &App, _env: &Environment) -> Result<Option<SetupPhase>> {
+        Ok(Some(SetupPhase::new(vec![
+            Pkg::new("leiningen"),
+            Pkg::new("jdk8"),
```
Milo123459, line 23: can the jdk version be customised? is it fixed at 8?

Milo123459 (changes requested):
This will need generate_plan_tests as well

src/providers/clojure.rs
AhmedMozaly: I've used JavaProvider as a reference, so this would be a common question about JVM providers
But worth raising!

AhmedMozaly (commented):
Add `generate_plan_tests`
@Milo123459

src/providers/clojure.rs
AhmedMozaly: Ok, I managed to add support for `NIXPACKS_JDK_VERSION` environment variable & `.jdk-version` file 
Once this is merged, we need to discuss adding to JavaProvider as well 

FYI, Nix has `jdk8`, `jdk11` and just `jdk` which has the most recent version I think

AhmedMozaly: Ok, now we support all available jdk nix packages with these values
- 8
- 11
- latest

AhmedMozaly (commented):
Cleanups

Milo123459:
Can you rebase this? I don't think we need all the Yarn cache changes.

src/providers/clojure.rs
```diff
@@ -0,0 +1,167 @@
...
+        // If not from configs, get it from the .jdk-version file
+        if custom_version.is_none() && app.includes_file(".jdk-version") {
+            custom_version = Some(app.read_file(".jdk-version")?);
+        }
+
+        // If it's still none, return default
+        if custom_version.is_none() {
+            return Ok(Pkg::new(DEFAULT_JDK_PKG_NAME));
+        }
+        let custom_version = custom_version.unwrap();
```
coffee-cup, line 55: Nit: This can be shortened to

```rs
        let custom_version = match custom_version {
            Some(v) => v,
            None => return Ok(Pkg::new(DEFAULT_JDK_PKG_NAME)),
        };
```

It also reads a bit nicer and doesn't have to unwrap

coffee-cup (dismissed):
One small comment but besides that it looks great!

src/providers/clojure.rs
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            matches.get(1)
+        };
+
+        let value = match matched_value {
+            Some(m) => m.as_str(),
+            None => "_",
+        };
+
+        return Ok(value.to_string());
```
JakeCooper, line 81: nit: should heed the GH action lints

src/providers/clojure.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn get_custom_version(app: &App, env: &Environment) -> Result<String> {
+        // Fetch version from configs
+        let mut custom_version = env.get_config_variable("JDK_VERSION");
+
+        // If not from configs, get it from the .jdk-version file
+        if custom_version.is_none() && app.includes_file(".jdk-version") {
+            custom_version = Some(app.read_file(".jdk-version")?);
+        }
+
+        return match custom_version {
```
coffee-cup, line 51: small lint here. same below.

image

coffee-cup (approved):

PR #343, opened 2022-07-21
do not run yarn berry with `--immutable`

coffee-cup:
## What does this PR address?
Yarn berry was seemingly updating the lockfile even after it was just generated. Not exactly sure why it is happening but since it can happen, we probably shouldn't be installing the deps with `--immutable`.


Milo123459:
Do you mean we probably should**n't** be installing deps with --imutable?

coffee-cup:
Oh yeah, typo. Updated the original

Milo123459 (dismissed):
LGTM

AhmedMozaly (approved):

tests/docker_run_tests.rs
```diff
@@ -253,6 +253,9 @@ fn test_yarn_custom_version() {
 fn test_yarn_berry() {
     let name = simple_build("./examples/node-yarn-berry");
     let output = run_image(name, None);
+
+    println!("Output: {}", output);
```
AhmedMozaly, line 257: Why we need to log output?

coffee-cup: We do not, I was just testing stuff. Good catch

AhmedMozaly (approved):

Milo123459 (approved):

PR #346, opened 2022-07-22
Sanitize both cache key and directory

AhmedMozaly:
While working on adding caching to Clojure provider, I've set the caching directory to `~/.m2` (Maven directory used by [Lein](https://leiningen.org/) for caching)

Got this caching path `DdFtDTR8ns0-~/.m2` which represents invalid path that lead to disabling caching entirely 
Sanitizing both cache key + cache dir would end up with something like `DdFtDTR8ns0-m2` which is safe & ensure caching won't be disturbed by dir values

### DIFF
--- a/src/nixpacks/cache.rs
+++ b/src/nixpacks/cache.rs
@@ -24,6 +24,6 @@ mod tests {
             sanitize_cache_key("s p a c e s".to_string()),
             "s-p-a-c-e-s".to_string()
         );
-        assert_eq!(sanitize_cache_key("~/.m2".to_string()), "-m2".to_string());
+        assert_eq!(sanitize_cache_key("/.m2".to_string()), "-m2".to_string());
     }
 }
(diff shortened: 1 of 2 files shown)


src/nixpacks/cache.rs
```diff
@@ -1,5 +1,16 @@
...
-    cache_key.replace(' ', "-")
+    cache_key
+        .chars()
+        .filter(|x| match x {
+            // remove these chars
+            '/' | '.' => false,
+            _ => true,
+        })
+        .map(|x| match x {
+            ' ' | '~' => '-',
```
coffee-cup, line 10: We should be able to safely map `~` to `/root`. That would allow cache directories to be something like `~/.cache/something`, which is fairly common in the unix world. At the moment users have nixpacks need to know that root is `/root`

AhmedMozaly: mapping `~` to `/root` I think should be for the target cache directory 
I've implemented & pushed please have a look, this is a sample result for one of my local runs

`CACHED [stage-0 6/8] RUN --mount=type=cache,id=DdFtDTR8ns0--root-m2,target=/root/.m2 `

coffee-cup (approved):

PR #347, opened 2022-07-22
Downgrade serde_with

coffee-cup:
For whatever reason, `serde_with` 2.0.0 will not build/install on my machine. I have a feeling it has to do with my arm mac and the version is not compatible. Since we are not using any `serde_with` 2.0.0 features, I've downgraded to the latest 1.X version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,9 +125,9 @@ dependencies = [
 
 [[package]]
 name = "darling"
-version = "0.14.1"
+version = "0.13.4"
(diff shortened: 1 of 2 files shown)


Milo123459:
Locally, can you try a `cargo clean` and `rm Cargo.lock`, then a `cargo build` to see if that fixes the issue?

coffee-cup:
> Locally, can you try a `cargo clean` and `rm Cargo.lock`, then a `cargo build` to see if that fixes the issue?

Same result :/

PR #349, opened 2022-07-22
Spike: Delete DOCKERFILE if found before writing to Dockerfile

AhmedMozaly:
By Spike: I mean we can decide not to merge this PR, it is meant for discussing proposed changes (instead of only describing it)

Currently, if Nixpacks build is called on a directory that already has a file named `DOCKERFILE` (or any other case than Dockerfile or dockerfile) It will fail with this error

```
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount938323349/Dockerfile: no such file or directory
 
Error: Docker build failed
```

So, better to make sure we're writing to the right path by removing any file with conflicting case 
Instead of users spending time figuring out why builds are failing with no clear error message about the issue

What do you think?

### DIFF
--- a/src/nixpacks/builder/docker.rs
+++ b/src/nixpacks/builder/docker.rs
@@ -143,6 +143,14 @@ impl DockerBuilder {
         let dockerfile = self.create_dockerfile(plan, env);
 
         let dockerfile_path = PathBuf::from(dest).join(PathBuf::from("Dockerfile"));
+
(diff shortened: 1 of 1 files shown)


AhmedMozaly:
We can still write a warning, but the question is: should this block our build process?

AhmedMozaly:
After some discussion, we will handle this issue with different approach
Closing this PR

PR #351, opened 2022-07-22
exclude examples and docs directories

coffee-cup:
We've hit the filesize limit when attempting to publish a new version to crates.io. This PR excludes the `examples/` and `docs/` directories since they do not need to be included.


JakeCooper (approved):

Milo123459:
Could we also use something like `cargo diet`?

AhmedMozaly (approved):

coffee-cup:
> Could we also use something like `cargo diet`? https://lib.rs/cargo-diet

image

Milo123459 (approved):

PR #355, opened 2022-07-22
Build plan snapshot testing

coffee-cup:
## What does this PR address?

Use [cargo-insta](https://github.com/mitsuhiko/insta) for snapshot testing the build plan tests. This is useful since it will tests all parts of the build plan for each tests as opposed to just the parts that get manually asserted. This also makes updating the build plan format a lot easier since we don't have to go and edit a 900 line file everytime.

I am not snapshot testing and build plans that rely on `ARCH`. This brings up a larger question in that we should be able to build for other platforms in the tests and not rely on the `ARCH` constant. But that should be fixed in another PR.


JakeCooper (approved):

Milo123459:
I'm not sure, this kind of looks painful to maintain.

tests/snapshots/generate_plan_tests__bun.snap
```diff
@@ -0,0 +1,39 @@
+---
+source: tests/generate_plan_tests.rs
+expression: plan
+---
+{
+  "version": "0.2.9",
```
Milo123459, line 6: Do we have to input the version every time for each test? I presume we have to update it on every new version as well?

coffee-cup:
> I'm not sure, this kind of looks painful to maintain.

Insta has tools to automatically review and update the snapshots. The snapshot files should never be edited manually.

This video explains how it works fairly well

https://www.youtube.com/watch?v=rCHrMqE4JOY

tests/snapshots/generate_plan_tests__bun.snap
coffee-cup: No these are never edited manually. They are used to compare the build plan output and make sure there are no changes, if there are changes you use `cargo insta review` to go through them one by one and either accept or reject the updates.

For example, after manually editing a `.snap` file and running

```
cargo insta test --review -- --test generate_plan_tests
```

I get

image

Pressing `a` will automatically update the snapshot with the changes

PR #368, opened 2022-07-26
Search All Python files Django + Support Double Quote identification

### LINKED ISSUE #367: Non-Trivial Django Applications Fail to Containerize

### Describe the bug

There are a few issues with the Django detection ATM

1) The regex uses single quotes only to parse for the WSGI application to pipe to the start command.

2) We only parse the settings.py file, but the WSGI application name can be in ANY of the imported settings files (e.g a folder)

### To reproduce

1) In settings.py, use single quotes instead of double quotes (e.g `WSGI_APPLICATION = "portfolio.wsgi.application"`)

2) Define the WSGI_APPLICATION in a folder called settings with a \_\_init\_\_.py file (using singlequotes)

### Expected behavior

Both should detect and build as Django

### Environment

OSX arm64 M1 Mac

JakeCooper:
## What does this PR address?
Fixes an issue where the WSGI_APPLICATION identified could be somewhere other than 
settings.py, as well as supports the usecase of double quotes, as well as the original single quotes usecase

Fixes #367 (issue)

## Before submitting:

- [x] Make sure to follow [GitHub's 
guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on creating PR.
- [x] Did you read through [contribution 
guidelines](https://github.com/railwayapp/nixpacks/blob/main/CONTRIBUTING.md)?
- [ ] Did you write tests to cover your changes? Did it pass locally when running `cargo test`?
- [ ] Have you run `cargo fmt`, `cargo check` and `cargo clippy` locally to ensure that CI passes?

### DIFF
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -196,7 +196,7 @@ impl PythonProvider {
 
     fn get_django_app_name(app: &App, _env: &Environment) -> Result<String> {
         // Look for the settings.py file
-        let paths = app.find_files("/**/settings.py").unwrap();
+        let paths = app.find_files("/*.py").unwrap();
 
         // Generate regex to find the application name
         let re = Regex::new(r"WSGI_APPLICATION = '(.*).application'").unwrap();
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -201,8 +201,6 @@ impl PythonProvider {
         // Generate regex to find the application name
         let re = Regex::new(r#"WSGI_APPLICATION = ["|'](.*).application["|']"#).unwrap();
 
-        println!("{:?}", paths.len());
-
         // Search all settings.py matches
         for path in paths {
             let path_buf = fs::canonicalize(path)?;
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -191,15 +191,17 @@ impl PythonProvider {
         // Check for the engine database type in settings.py
         let re = Regex::new(r"django.db.backends.postgresql").unwrap();
 
-        app.find_match(&re, "/**/settings.py")
+        app.find_match(&re, "/**/*.py")
     }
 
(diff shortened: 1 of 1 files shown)


src/providers/python.rs
```diff
@@ -196,7 +196,7 @@ impl PythonProvider {
 
     fn get_django_app_name(app: &App, _env: &Environment) -> Result<String> {
         // Look for the settings.py file
-        let paths = app.find_files("/**/settings.py").unwrap();
+        let paths = app.find_files("/*.py").unwrap();
```
Milo123459, line 199:
```suggestion
        let paths = app.find_files("/**/*.py").unwrap();
```

JakeCooper: Shouldn't matter should it?

Milo123459: /*.py only matches python files in the top-level directory, that glob matches all python files in all sub-directories.

src/providers/python.rs
```diff
@@ -191,15 +191,17 @@ impl PythonProvider {
...
     fn get_django_app_name(app: &App, _env: &Environment) -> Result<String> {
         // Look for the settings.py file
-        let paths = app.find_files("/**/settings.py").unwrap();
+        let paths = app.find_files("/**/*.py").unwrap();
 
         // Generate regex to find the application name
-        let re = Regex::new(r"WSGI_APPLICATION = '(.*).application'").unwrap();
+        let re = Regex::new(r#"WSGI_APPLICATION = ["|'](.*).application["|']"#).unwrap();
+
+        println!("{:?}", paths.len());
```
coffee-cup, line 204: Can remove this

coffee-cup (approved):

PR #376, opened 2022-07-31
Erlang support

marianoguerra:
## What does this PR address?

Adds initial support for Erlang

If there's a way to test deploying to railwayapp before merging here I can do that but I don't know how.

Note: for the start task I need to know the name of the release, which is in the rebar.config file which has an Erlang specific format called "erlang terms" (a subset of the Erlang syntax), the simplest way that would handle it correctly other than regexes is to add an erlang lexer library and extract the release name from the lexer stream, let me know if I should use a Regex instead (which may not work in some cases).

Another solution for this would be to run an Erlang one liner to parse the file and get the release name from the data, but that would require to have Erlang available at build time (with something like nix-shell) but I think it' not possible yet.

This will also be required for Elixir (which I plan to add) but even worse because mix.exs in Elixir is Elixir code that has to run to get the release name, so parsing may not be enough.

## Before submitting:


marianoguerra:
I rebased my changes on main some days ago and I see some changes in other files appear in my PR, not sure what to do about them.

src/providers/erlang.rs
```diff
@@ -0,0 +1,91 @@
...
+            Pkg::new("rebar3"),
+        ])))
+    }
+
+    fn install(&self, _app: &App, _env: &Environment) -> Result<Option<InstallPhase>> {
+        Ok(Some(InstallPhase::new("rebar3 get-deps".into())))
+    }
+
+    fn build(&self, _app: &App, _env: &Environment) -> Result<Option<BuildPhase>> {
+        Ok(Some(BuildPhase::new("rebar3 release".into())))
```
AhmedMozaly, line 43: Does adding `as prod` would make any difference in build output?

src/providers/erlang.rs
```diff
@@ -0,0 +1,91 @@
...
+                get_release_name_from_rebar_config(&rebar_file.to_string_lossy())
+            {
+                Ok(Some(StartPhase::new(format!(
+                    "./_build/default/rel/{}/bin/{} foreground",
+                    name, name
+                ))))
+            } else {
+                Err(anyhow::anyhow!(
+                    "Couldn't find release name in rebar.config"
+                ))
```
coffee-cup, line 58: nit: Can you the `bail!` macro here instead

src/providers/erlang.rs
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+            Err(anyhow::anyhow!(
+                "Couldn't find release name in rebar.config"
+            ))
+        }
+    }
+}
+
+fn get_release_name_from_rebar_config(path: &str) -> Result<Option<String>, Error> {
+    let mut file = File::open(path)?;
```
coffee-cup, line 69: This shouldn't use `File` directly and should instead interact with the `App`. `app.read_file` will work here

coffee-cup (commented):
A few comments but overall looks good. Thanks for PRing this 😄

marianoguerra:
should I rebase main myself or you will do it?

Milo123459:
could you do it please?

marianoguerra:
I created a new PR from a different branch because I made a mess with this branch https://github.com/railwayapp/nixpacks/pull/421

PR #380, opened 2022-08-02
Write nixpacks related files to .nixpacks directory

AhmedMozaly:
## What does this PR address?
This PR change the output path for Nixpacks-related files (like Dockerfile & environment.nix) to be created under `.nixpacks` sub-directory instead of the root path of the output dir.

This should avoid the need to override certain user files (like Dockerfile) if was already exists and keep away some unexpected behavior, Like finding out DOCKERFILE in user provided code instead of Dockerfile which cause docker build step to break later.

## Before submitting:

### DIFF
--- a/src/nixpacks/builder/docker.rs
+++ b/src/nixpacks/builder/docker.rs
@@ -20,6 +20,7 @@ const DOT_NIXPACKS_DIR: &'static &str = &".nixpacks";
 struct OutputDir {
     root_path: PathBuf,
     dockerfile_path: PathBuf,
+    environment_nix_path: PathBuf,
 }
 
 impl OutputDir {
@@ -34,19 +35,14 @@ impl OutputDir {
         }
 
         let dockerfile_path = PathBuf::from(&dot_nixpacks_dir).join(PathBuf::from("Dockerfile"));
+        let environment_nix_path =
+            PathBuf::from(&dot_nixpacks_dir).join(PathBuf::from("environment.nix"));
 
         Ok(OutputDir {
             root_path,
             dockerfile_path,
+            environment_nix_path,
         })
-    }
-
-    pub fn get_nixpacks_file_path(&self, file_name: String) -> String {
-        PathBuf::from(&self.root_path)
-            .join(PathBuf::from(DOT_NIXPACKS_DIR))
-            .join(PathBuf::from(file_name))
-            .display()
-            .to_string()
     }
 }
 
@@ -99,7 +95,7 @@ impl Builder for DockerBuilder {
         self.write_assets(plan, dest).context("Writing assets")?;
         self.write_dockerfile(plan, &output_dir.dockerfile_path, env)
             .context("Writing Dockerfile")?;
-        self.write_nix_expression(plan, &output_dir)
+        self.write_nix_expression(plan, &output_dir.environment_nix_path)
             .context("Writing NIx expression")?;
 
         // Only build if the --out flag was not specified
@@ -201,8 +197,7 @@ impl DockerBuilder {
         Ok(())
     }
 
-    fn write_nix_expression(&self, plan: &BuildPlan, output_dir: &OutputDir) -> Result<()> {
-        let environment_nix_path = output_dir.get_nixpacks_file_path("environment.nix".to_string());
+    fn write_nix_expression(&self, plan: &BuildPlan, environment_nix_path: &PathBuf) -> Result<()> {
         let nix_expression = nix::create_nix_expression(plan);
 
         let mut nix_file =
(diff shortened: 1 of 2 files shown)


src/nixpacks/builder/docker.rs
```diff
@@ -139,21 +204,25 @@ impl DockerBuilder {
...
-        let dockerfile = self.create_dockerfile(plan, env);
-
-        let dockerfile_path = PathBuf::from(dest).join(PathBuf::from("Dockerfile"));
-        File::create(dockerfile_path.clone()).context("Creating Dockerfile file")?;
+    fn write_dockerfile(
+        &self,
+        plan: &BuildPlan,
+        dockerfile_path: &str,
+        env: &Environment,
+        environment_nix_path: &str,
```
coffee-cup, line 212: It is a bit strange to be passing in the nix environment path when creating the dockerfile. I think it would be better to just pass in the full path of the dockerfile that should be created.

src/nixpacks/builder/docker.rs
```diff
@@ -15,6 +15,50 @@ use indoc::formatdoc;
...
+
+        let environment_nix_relative_path = PathBuf::from(DOT_NIXPACKS_DIR)
+            .join(PathBuf::from("environment.nix"))
+            .display()
+            .to_string();
+
+        let environment_nix_absolute_path = PathBuf::from(root_path)
+            .join(PathBuf::from(&environment_nix_relative_path))
+            .display()
+            .to_string();
```
coffee-cup, line 47: Instead of storing both the absolute and relative paths for this single file, how about storing JUST either the absolute or relative paths for both the dockerfile and environment file.

src/nixpacks/builder/docker.rs
```diff
@@ -15,6 +15,50 @@ use indoc::formatdoc;
 use tempdir::TempDir;
 use uuid::Uuid;
 
+const DOT_NIXPACKS_DIR: &'static &str = &".nixpacks";
+
+struct OutputDir {
+    root_path: String,
+    dockerfile_path: String,
+    environment_nix_relative_path: String,
+    environment_nix_absolute_path: String,
```
coffee-cup, line 24: Any reason for storing these as strings instead of `PathBuf`s?

src/nixpacks/builder/docker.rs
```diff
@@ -60,16 +107,23 @@ impl Builder for DockerBuilder {
...
+        self.write_nix_expression(plan, &output_dir.environment_nix_absolute_path)
             .context("Writing NIx expression")?;
 
         // Only build if the --out flag was not specified
         if self.options.out_dir.is_none() {
-            let mut docker_build_cmd = self.get_docker_build_cmd(plan, name.as_str(), dest)?;
+            let mut docker_build_cmd =
+                self.get_docker_build_cmd(plan, name.as_str(), &output_dir)?;
+            let answer = format!("{:?}", docker_build_cmd);
+            println!("command is {}", answer); // Execute docker build;
```
coffee-cup, line 125: Can remove this

coffee-cup (commented):
Looks good, just a few comments

src/nixpacks/builder/docker.rs
AhmedMozaly: I've removed both & added new `get_nixpacks_file_path` method to resolve absolute path

src/nixpacks/builder/docker.rs
```diff
@@ -15,6 +15,38 @@ use indoc::formatdoc;
...
+        let dot_nixpacks_dir = PathBuf::from(&root_path)
+            .join(PathBuf::from(DOT_NIXPACKS_DIR))
+            .display()
+            .to_string();
+
+        if fs::metadata(&dot_nixpacks_dir).is_err() {
+            fs::create_dir_all(&dot_nixpacks_dir)?;
+        }
+
+        let dockerfile_path = PathBuf::from(&dot_nixpacks_dir).join(PathBuf::from("Dockerfile"));
```
coffee-cup, line 38: nit: Move the `"Dockerfile"` variable to a constant too

src/nixpacks/builder/docker.rs
```diff
@@ -222,7 +276,7 @@ impl DockerBuilder {
         };
 
         // -- Setup
-        let mut setup_files: Vec<String> = vec!["environment.nix".to_string()];
+        let mut setup_files: Vec<String> = vec![environment_nix_path];
```
coffee-cup, line 279: I think this can just use `ENVIRONMENT_NIX` instead of `environment_nix_path`

coffee-cup (approved):
Some minor comments but otherwise good 😄

src/nixpacks/builder/docker.rs
AhmedMozaly: Here, we need the relative path `.nixpacks/environment.nix` thats why we need to compute it first. because ENVIRONMENT_NIX only represents `environment.nix`

PR #381, opened 2022-08-02
Remove node version log

coffee-cup:


Milo123459:
Not sure I think this is a good idea. It's good to communicate over that no version could be matched.

coffee-cup:
> Not sure I think this is a good idea. It's good to communicate over that no version could be matched.

Nothing else prints the version number. It just feels weird and out of place. If we want to log we should output it in the build plan or log for everything. This is also a println in the function `version_number_to_pkg` which is a utility that function that should not be globally logging.

Milo123459:
Could we add a way to add fields to the initial box that is shown? We could add a notes section and add this in.

Milo123459 (approved):

coffee-cup:
The box is derived from the build plan. If we want to show additional information on the generation of the build plan, we should setup a proper logging system (i.e. debug/info/warn/error). This would probably get output before the build plan.

PR #387, opened 2022-08-07
feat: suppport for NX monorepos

FinnDore:
## What does this PR address?
Out of the box support for [NX mono repos](https://nx.dev/). ( nexjs apps, express apps and apps that build to an index.js )

- [X] Make sure to follow [GitHub's guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on creating PR.
- [x] Did your changes require updates to the documentation? Have you updated those accordingly? - New env var so yes?


src/providers/node.rs
```diff
@@ -83,9 +120,19 @@ impl Provider for NodeProvider {
...
+        if NodeProvider::is_nx_monorepo(app) {
+            let app_name = NodeProvider::get_nx_app_name(app, _env)?.unwrap();
+            build_phase.add_cmd(format!(
+                "{} run build {} --configuration=production",
+                pkg_manager, app_name
+            ));
+            // Remove the package.json so when we run node /dist/**/file.js node does
+            // not look at the package.json and infer things it should not, Such as wether its a es module or not.
+            build_phase.add_cmd("rm /app/package.json".to_owned());
+            build_phase.add_cache_directory(NodeProvider::get_nx_output_path(app, _env).unwrap());
```
FinnDore, line 134: Should I be caching the output folder of the build?

FinnDore:
Think this is ready for a review whenever someone is free 😊

examples/node-nx/.dockerignore
```diff
@@ -0,0 +1 @@
+node_modules
```
coffee-cup, line 1: This file isn't needed

examples/node-nx/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
```
coffee-cup, line 1: This file isn't needed either

src/providers/node.rs
coffee-cup: The cached directories do not appear in the final image so it wouldn't work to cache the build output directory.

src/providers/node.rs
```diff
@@ -83,9 +120,19 @@ impl Provider for NodeProvider {
...
+
+        if NodeProvider::is_nx_monorepo(app) {
+            let app_name = NodeProvider::get_nx_app_name(app, _env)?.unwrap();
+            build_phase.add_cmd(format!(
+                "{} run build {} --configuration=production",
+                pkg_manager, app_name
+            ));
+            // Remove the package.json so when we run node /dist/**/file.js node does
+            // not look at the package.json and infer things it should not, Such as wether its a es module or not.
+            build_phase.add_cmd("rm /app/package.json".to_owned());
```
coffee-cup, line 133: I am a little confused by this. It is not common to remove the package.json file. Why is it necessary in this case? I've run many NX apps without the need for removing this file. I also know a few apps that rely on this app existing in production.

This also won't work if the start command is overriden to be something like `npm run web:start`

src/providers/node.rs
```diff
@@ -143,7 +190,30 @@ impl NodeProvider {
...
     }
 
-    pub fn get_start_cmd(app: &App) -> Result<Option<String>> {
+    pub fn get_start_cmd(app: &App, env: &Environment) -> Result<Option<String>> {
+        if NodeProvider::is_nx_monorepo(app) {
+            let output_path = NodeProvider::get_nx_output_path(app, env)?;
+            let project_json = NodeProvider::get_nx_project_json_for_app(app, env)?;
+
+            let executor = project_json.targets.build.executor;
+            if executor == "@nrwl/next:build" {
```
coffee-cup, line 199: A few comments for the NX start command

- The port shouldn't be set when running a Next app. It should just use the default
- Instead of cd'ing into the directory, I think it is cleaner to start the app with Nx. E.g. `nx run frontend:start`
- The default start command for all NX apps should be running the start command in the app (if it exists). For example for something like Yarn/NX monorepos, each sub-app will have it's own start command that we should use (if it exists). Then we can fallback to running `dist/main.js`

src/providers/node.rs
```diff
@@ -367,6 +437,59 @@ impl NodeProvider {
...
+            return Ok(Some(app_name));
+        }
+
+        if let Ok(nx_json) = app.read_json::<NxJson>("nx.json") {
+            if let Some(default_project) = nx_json.default_project {
+                return Ok(Some(default_project.as_str().unwrap().to_owned()));
+            }
+        }
+
+        return Err(anyhow!("Could not derive nx app to build and run. Please add a default project to your nx config or set NIXPACKS_{}", NX_APP_NAME_ENV_VAR));
```
coffee-cup, line 456: Tip: use the `bail!` macro here as a shorthand for `return Err(anyhow!`

src/providers/node.rs
```diff
@@ -367,6 +437,59 @@ impl NodeProvider {
...
+    }
+
+    pub fn get_nx_project_json_for_app(app: &App, env: &Environment) -> Result<ProjectJson> {
+        let app_name = NodeProvider::get_nx_app_name(app, env)?.unwrap();
+        let project_path = format!("./apps/{}/project.json", app_name);
+        let nx_app_project_json = app.read_json::<ProjectJson>(&project_path);
+
+        if nx_app_project_json.is_err() {
+            return Err(anyhow!(format!(
```
coffee-cup, line 465: Correct me if I am wrong, but I don't think that a `project.json` file is required for all NX apps and the config for each app can be at the project root. In this case I think it makes more sense for this function to return `Result<Option<ProjectJson>>` instead of throwing an error.

src/providers/node.rs
```diff
@@ -367,6 +437,59 @@ impl NodeProvider {
...
+
+    pub fn get_nx_output_path(app: &App, env: &Environment) -> Result<String> {
+        let project_json = NodeProvider::get_nx_project_json_for_app(app, env)?;
+        if project_json.targets.build.options.output_path.is_some() {
+            Ok(project_json
+                .targets
+                .build
+                .options
+                .output_path
+                .unwrap()
```
coffee-cup, line 482: Please no unwrapping. Lets gracefully handle the case where the output path is not set properly.

Same with the unwrap below.

src/providers/node.rs
FinnDore: The reason I chose not to use the `serve` target is that it appears to be for development purposes as it includes a websocket for use with a  debugger and hot-reload on file change. If this is ok we can totally use `nx run <app-name>:serve:production` see the output below. The other three targets that are generated by the default express generator (`@nrwl/express:aplication`) are `lint`, `build` and `test`.

```
npx nx run express-app:serve:production

> nx run express-app:serve:production

chunk (runtime: main) main.js (main) 564 bytes [entry] [rendered]
webpack compiled successfully (02f31e3162b7080a)
Debugger listening on ws://localhost:9229/666612ae-1046-4d7f-aefd-f39ef9b3fc16
Debugger listening on ws://localhost:9229/666612ae-1046-4d7f-aefd-f39ef9b3fc16
For help, see: https://nodejs.org/en/docs/inspector
nx express app works
chunk (runtime: main) main.js (main) 572 bytes [entry] [rendered]
webpack compiled successfully (4275d6c19665a755)
Command failed: taskkill /pid 28756 /T /F
ERROR: The process "28756" not found.

Debugger listening on ws://localhost:9229/f9432e74-f926-4596-8ad1-a6a5175ef466
Debugger listening on ws://localhost:9229/f9432e74-f926-4596-8ad1-a6a5175ef466
For help, see: https://nodejs.org/en/docs/inspector
nx express app works, reload
```

( same story with `@nrwl/node:webpack` )

The reason for deleting the `package.json` was to solve an issue where running `node <path-to-app>/main.js` would have some interaction with the `package.json`. This iirc. Maybe it would be better ( if we don't use the serve target ) to cd into the output folder and run `node main.js`?

```
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/app/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

examples/node-nx/.editorconfig
FinnDore: Removed

src/providers/node.rs
FinnDore: Oh neat 😊! done this.

src/providers/node.rs
FinnDore: True! It would fail way before they even try to build with nixpacks.

src/providers/node.rs
FinnDore: Good shout, lmk if the changes are good. It certainly looks more readable.

examples/node-nx/.dockerignore
FinnDore: Removed

src/providers/node.rs
FinnDore: Makes sense, deleted this

src/providers/node.rs
FinnDore: Same as the other start commands.

Regarding the port.  The default port is something like 3200, I would expect a website to be hosted on port 80 and without this change, it would not be. removing this the user would have to overnight the start command manually to use port 80. 

Would the default port be acceptable?

src/providers/node.rs
coffee-cup: I agree it doesn't make sense to use the `serve` target. But lots of apps have another separate `start` task/script that is meant for production. If that target is found we can run that command. As for the ES module error, I think that can be solved with a field in the package.json file. If the app was deployed and run in other production environments it would not be common to delete the `package.json` file.

Unfortunately deleting this file is not really feasible with Nixpacks since some users may override the start command with the `NIXPACKS_START_CMD` variable to something like `npm run start`. This should work even for NX apps.

src/providers/node.rs
coffee-cup: For Next apps the port is overridable with the `PORT` environment variable and that is common in many other frameworks. Also, when starting the container, you can map port 3200 inside of the container to any other port on the outside, so the default port in the container does not matter.

tests/docker_run_tests.rs
```diff
@@ -228,6 +228,78 @@ fn test_node() {
     assert!(run_image(name, None).contains("Hello from Node"));
 }
 
+#[test]
+fn test_node_nx_next() {
+    let n = create_network();
+    let config = Config {
+        environment_variables: EnvironmentVariables::from([]),
+        network: Some(n.name.to_string()),
```
coffee-cup, line 236: I believe that creating a network here isn't necessary since the container does not interact with any other containers (so it can be `None`).

src/providers/node.rs
FinnDore: Sounds good to me. I'll get on this change when I can.

src/providers/node.rs
coffee-cup: +1, I agree that start command priority order makes sense.

As for editing package.json, I think it is fairly common to do when running generated code in a production environment. I must admit, I haven't written raw JavaScript in a long time (mainly just TS), but changing the module type in package.json should be okay. The main thing is that we should not be manually deleting these files for the user just to run their code.

tests/docker_run_tests.rs
FinnDore: neat, done this now.

src/providers/node.rs
FinnDore: That's fair, ill get on that tomorrow (hopefully) when I have time, serde should make this easy. Thanks for the insights as well.

src/providers/node.rs
FinnDore: Should be done now. Just the run commands to change as of present.

src/providers/node.rs
```diff
@@ -34,6 +40,37 @@ pub struct PackageJson {
...
+
+#[derive(Debug, Serialize, PartialEq, Deserialize)]
+pub struct Targets {
+    build: Build,
+}
+
+#[derive(Debug, Serialize, PartialEq, Deserialize)]
+pub struct ProjectJson {
+    targets: Targets,
+}
```
coffee-cup, line 72: Can this be moved into an NX module? It isn't clear that these are just for NX monorepos. I think we should have

```
providers/
  node/
    mod.rs
    nx.rs
```

Can use `git mv providers/node.rs providers/node/mod.rs` to cleanly move it over.

FinnDore: Agreed. That git command is neat as well, it worked great once I realised I needed to create the node folder manually 😄.

src/providers/node.rs
FinnDore: Just removed the `rm package.json` step and ran the tests ran fine, so I think I've been lying about the module-type issue. 
Ill make the start command change and then I can asses if its really an issue

JakeCooper:
As a result of this provider, can we revert #382?

FinnDore:
Tbh I have never needed this env var on Railway. Having said that @ChristianIvicevic said not having it causes a sigTerm? I have noticed a sig term on vercel before when using nx but it's very flakey and may be caused by me not setting this env var.

ChristianIvicevic:
`CI=true` makes a lot of sense with Node projects such as disabling watch modes for tools (such as tests) or in the case of Nx disabling the background daemon that caused SIGTERMs.

coffee-cup:
> As a result of this provider, can we revert #382?
> 
> The reason we added it was for the nwl stuff

This PR does not change the fact that `CI=true` is needed. It just adds better first class support for NX apps.

src/providers/node/mod.rs
```diff
@@ -83,9 +96,12 @@ impl Provider for NodeProvider {
 
     fn build(&self, app: &App, _env: &Environment) -> Result<Option<BuildPhase>> {
         let mut build_phase = BuildPhase::default();
+        let pkg_manager = NodeProvider::get_package_manager(app);
 
-        if NodeProvider::has_script(app, "build")? {
-            let pkg_manager = NodeProvider::get_package_manager(app);
+        if NodeProvider::is_nx_monorepo(app) {
+            let app_name = NodeProvider::get_nx_app_name(app, _env)?.unwrap();
+            build_phase.add_cmd(format!("{} run {}:build:production", pkg_manager, app_name));
```
coffee-cup, line 103: Should this be `npm nx run APP:build:production` since this script may not exist. I am seeing this error when building the default NX app.

image

FinnDore: I believe you are correct. I'm refactoring the build command so it works no matter what package manager the user is using. Just didn't get enough time last night  😆

FinnDore:
Think all the ducks are in order now. This should be ready for a re-review whenever you are free next week or so 😀.

coffee-cup (approved):
It's looking good! Can you please fix the linting errors and then I will merge :D

FinnDore:
Done!

FinnDore:
Odd, somehow I missed some. I'll get on that asap tomorrow.

coffee-cup:
I just went ahead and fixed the remaining lint errors (Literally just had to run `cargo clippy --fix`) 😄

PR #394, opened 2022-08-10
Add Nixpacks builder environment variable name

JakeCooper:
## What does this PR address?

Injects meta information variables for containers to determine which provider is participating

## Before submitting:


AhmedMozaly (commented):
Could we extend this a bit more with something like `NIXPACKS_LABELS` that can hold values like `python, django`

This could help us maintain the info captured during Nixpacks `detect` phase

Which could be used to provide even better Developer-Experience by offering more helpful hints for common issues (Like PORT not configured)

AhmedMozaly:
I already have a Todo to raise a discussion about refactoring `detect`, currently it returns only boolean

I think adding list of strings for detected facts/labels can simply other phases & write these detected labels to environment variables (like this current PR)

coffee-cup (changes requested):
I don't think this should be the responsibility of each provider to add this environment variable. Each provider has a name so we should be able to insert that into the environment variables for all builds.

PR #402, opened 2022-08-14
Support cargo workspaces

### LINKED ISSUE #256: Support Rust workspace Cargo.toml
Tried this under Railway, but when your `Cargo.toml` is a workspace one, it doesn't work. It however should be allowed. Building a workspace will build all members, and while that may be not ideal, starting a binary will still work with it since it'll get generated. A person can also provide a custom build command targeting only one package when that's undesired behavior,  
`cargo build --release -p package-name`

```
Error: Generating start phase
 
Caused by:
 
0: Reading Cargo.toml
1: missing field `package` at line 1 column 1
```

devtomio:
## What does this PR address?

Adds support for Cargo [Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)

Fixes #256 

## Before submitting:


docs/pages/docs/providers/rust.md
```diff
@@ -46,3 +46,9 @@ These directories are cached between builds
 - Build: `~/.cargo/git`
 - Build: `~/.cargo/registry`
 - Build: `target`
+
+## [Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)
+
+Nixpacks will auto-detect if you are using Cargo Workspaces.
```
Milo123459, line 52: This should probably state that it respects `default_members`.

devtomio: Done in https://github.com/railwayapp/nixpacks/pull/402/commits/d853c9a1af7d579e217df2363b481395cc70b193

src/providers/rust.rs
```diff
@@ -233,6 +218,141 @@ impl RustProvider {
...
+                        app.read_toml::<Manifest>(&format!("{}/Cargo.toml", default_member))?;
+
+                    manifest.complete_from_path(
+                        &app.source.join(format!("{}/Cargo.toml", default_member)),
+                    )?;
+
+                    if let Some(package) = manifest.package {
+                        if !manifest.bin.is_empty() || manifest.lib.is_none() {
+                            return Ok(Some(package.name));
+                        }
```
coffee-cup, line 264: This bit of code is reused a few places. Can it be extracted into a function on the manifest. Something like `manifest.is_empty`

src/providers/rust.rs
```diff
@@ -233,6 +218,141 @@ impl RustProvider {
...
+                        .complete_from_path(&app.source.join(format!("{}/Cargo.toml", member)))?;
+
+                    if let Some(package) = manifest.package {
+                        if !manifest.bin.is_empty() || manifest.lib.is_none() {
+                            return Ok(Some(package.name));
+                        }
+                    }
+                }
+            }
+        }
```
coffee-cup, line 304: Sam with this code. It is reused from the default member stuff above. It could be abstracted into a function to remove the duplication.

devtomio: Done in https://github.com/railwayapp/nixpacks/pull/402/commits/7879e9f935bb57dabfa038b25ab6d6679df2d2cb

src/providers/rust.rs
devtomio: We can't extend an implementation of a struct unless we define an extension `trait`, which would be more harder.

coffee-cup (approved):
Looks good! Thanks for the PR 😄

docs/pages/docs/providers/rust.md
```diff
@@ -46,3 +46,12 @@ These directories are cached between builds
...
+This checks `workspace.default_members` first and then `workspace.members`.
+It also respects the `workspace.exclude` field.
+
+To set which workspace Nixpacks will build, just set the `NIXPACKS_CARGO_WORKSPACE`
```
prideout, line 56: I think this is actually specifying the _crate_ within the workspace, not the workspace.  Seems like an awkward name for the environment variable.

PR #403, opened 2022-08-14
Add a macro to help with plan tests

devtomio:
## What does this PR address?

Creates a separate helper crate to remove code duplication on plan tests. 

## Before submitting:


coffee-cup (approved):
Nice! I like this change. Eventually, when we have a file based config format (planned), we can use that instead of `test.env` files. But this works well for now.

PR #412, opened 2022-08-17
Remove Random PHP Folder

JakeCooper:
Don't think this should be here


JakeCooper:
Oh lord yup these are required

PR #415, opened 2022-08-17
Add node_modules/.bin to path

### LINKED ISSUE #399: Nixpacks does not make the installed node_modules executables available

### Describe the bug

When building a node project with custom commands (i.e.: `--build-cmd`), the installed npm packages are not available to be used in the image.

For example, trying to build a Blitz project and passing `blitz build` to `--build-cmd` fails because `blitz` is not found. 
Note that in a simple project, the build works without a custom command, but the point is that I'd expect whatever was installed to be available in the image. 

### To reproduce

1. Clone this repo: https://github.com/crondinini/blitzjs
2. Run nixpacks with: `nixpacks . build --build-cmd 'blitz dev'`
3. See the error:
```
[+] Building 36.6s (11/12)
 => [internal] load build definition from Dockerfile                                0.2s
 => => transferring dockerfile: 626B                                                0.1s
 => [internal] load .dockerignore                                                   0.2s
 => => transferring context: 2B                                                     0.0s
 => [internal] load metadata for ghcr.io/railwayapp/nixpacks:debian-1657820962      0.6s
 => [internal] load build context                                                   0.1s
 => => transferring context: 681.65kB                                               0.1s
 => [stage-0 1/8] FROM ghcr.io/railwayapp/nixpacks:debian-1657820962@sha256:f116f7  0.0s
 => CACHED [stage-0 2/8] WORKDIR /app/                                              0.0s
 => CACHED [stage-0 3/8] COPY environment.nix /app/                                 0.0s
 => CACHED [stage-0 4/8] RUN nix-env -if environment.nix                            0.0s
 => [stage-0 5/8] COPY . /app/                                                      0.1s
 => [stage-0 6/8] RUN --mount=type=cache,id=mgAt8pKUNJE-/usr/local/share/cache/ya  35.0s
 => ERROR [stage-0 7/8] RUN --mount=type=cache,id=mgAt8pKUNJE-node_modules/cache,t  0.4s
------
 > [stage-0 7/8] RUN --mount=type=cache,id=mgAt8pKUNJE-node_modules/cache,target=node_modules/.cache blitz dev:
#11 0.412 /bin/bash: line 1: blitz: command not found
```

### Expected behavior

I'd expect the installed packages to be available by adding `node_modules/.bin` to the PATH.

Looking at the code for other providers, both [`python`](https://github.com/railwayapp/nixpacks/blob/main/src/providers/python.rs#L73) and [`ruby`](https://github.com/railwayapp/nixpacks/blob/main/src/providers/ruby.rs#L57-L59) have a step during the install process that adds what looks like the `bin` folder related to the packages installed (I'm assuming this based on the folder name, but I could be wrong as I'm not familiar with those languages).

### Environment

MacOS Monterey 12.4
nixpacks 0.2.11

coffee-cup:
Fixes #399

Add the `./node_modules/.bin` directory to the path in Node apps.

### DIFF
--- a/src/providers/node/mod.rs
+++ b/src/providers/node/mod.rs
@@ -71,6 +71,7 @@ impl Provider for NodeProvider {
         // Install
         let mut install = Phase::install(Some(NodeProvider::get_install_command(app)));
         install.add_cache_directory(NodeProvider::get_package_manager_cache_dir(app));
+        install.add_path("/app/node_modules/.bin".to_string());
 
         // Cypress cache directory
         let all_deps = NodeProvider::get_all_deps(app)?;
@@ -79,7 +80,6 @@ impl Provider for NodeProvider {
         }
 
         // Build
-
         let mut build = if NodeProvider::is_nx_monorepo(app) {
             let app_name = NodeProvider::get_nx_app_name(app, env)?.unwrap();
             Phase::build(Some(format!("npx nx run {}:build:production", app_name)))
--- a/tests/snapshots/generate_plan_tests__node.snap
+++ b/tests/snapshots/generate_plan_tests__node.snap
@@ -21,6 +21,9 @@ expression: plan
       ],
       "cacheDirectories": [
         "/root/.npm"
+      ],
+      "paths": [
+        "/app/node_modules/.bin"
       ]
     },
     {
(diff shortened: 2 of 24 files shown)


Milo123459 (approved):

devtomio:
IMO it would be better to just make the files in `node_modules/.bin` executable instead of adding them to path to avoid name conflicts.

coffee-cup:
> IMO it would be better to just make the files in `node_modules/.bin` executable instead of adding them to path to avoid name conflicts.

These files are already executable. However, they are not available globally (e.g. running `prisma` wont' work) which is a fairly common thing to do in the Node world.

PR #418, opened 2022-08-18
Elixir support

JakeCooper:
This PR adds support for Elixir Phoenix


src/providers/elixir.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct DenoTasks {
+    pub start: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct DenoJson {
+    pub tasks: Option<DenoTasks>,
+}
```
coffee-cup, line 22: These are not needed

src/providers/elixir.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn name(&self) -> &str {
+        "elixir"
+    }
+
+    fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
+        Ok(app.includes_file("mix.exs"))
+    }
+
+    fn get_build_plan(&self, _app: &App, _env: &Environment) -> Result<Option<BuildPlan>> {
+        let mut plan = BuildPlan::default();
```
coffee-cup, line 36: bit nit: The plan can be created with all the phases at the bottom of this function with

```rs
let plan = BuildPlan::new(vec![setup_phase, install_phase, build_phase], Some(start_phase));
```

It does the same thing but it a few lines shorter.

coffee-cup (approved):

src/providers/elixir.rs
JakeCooper: Tried to do that but kept getting a borrow exception

coffee-cup (approved):

PR #421, opened 2022-08-19
Erlang Provider

marianoguerra:
## What does this PR address?

Adds initial support for Erlang

If there's a way to test deploying to railwayapp before merging here I can do that but I don't know how.

Note: for the start task I need to know the name of the release, which is in the rebar.config file which has an Erlang specific format called "erlang terms" (a subset of the Erlang syntax), the simplest way that would handle it correctly other than regexes is to add an erlang lexer library and extract the release name from the lexer stream, let me know if I should use a Regex instead (which may not work in some cases).

Another solution for this would be to run an Erlang one liner to parse the file and get the release name from the data, but that would require to have Erlang available at build time (with something like nix-shell) but I think it' not possible yet.

This will also be required for Elixir (which I plan to add) but even worse because mix.exs in Elixir is Elixir code that has to run to get the release name, so parsing may not be enough.

## Before submitting:


Milo123459:
Hey @marianoguerra - could you give us a status update on this?

marianoguerra:
I can't run all tests since they fill my disk, I run erlang ones and they pass locally, it's hard to stay up to date on main, my lastest commits made a mess, closing this

PR #422, opened 2022-08-19
Use a macro for docker tests

devtomio:
Utilizes the `test-helper` crate to generate docker tests.

Unrelated:
- Adds the [`mitsuhiko.insta`](https://marketplace.visualstudio.com/items?itemName=mitsuhiko.insta) extension to the devcontainer setup to enable code highlighting to snapshot files.


test-helper/src/lib.rs
```diff
@@ -99,3 +108,305 @@ pub fn generate_plan_tests(_tokens: TokenStream) -> TokenStream {
         .collect::<proc_macro2::TokenStream>()
         .into()
 }
+
+#[proc_macro]
+pub fn generate_docker_tests(_tokens: TokenStream) -> TokenStream {
```
coffee-cup, line 113: I am unsure about this change. I think it makes editing how the tests are run a lot less clear and means we can't easily add custom behaviour for a single test if we want to in the future. Each example also needs to be manually edited anyway to account for the test specific variables in `test.env`, which isn't that different then just editing `docker_run_tests`.

CONTRIBUTING.md
```diff
@@ -47,6 +47,25 @@ The snapshots are checked into CI and are reviewed as part of the PR. They ensur
...
+- `TEST_PLAN_CUSTOM_PKGS` - Sets custom nix packages.
+- `TEST_PLAN_ENVS` - Sets custom environment variables.
+
+### Docker Tests
+
+- `TEST_DOCKER_ENVS` - Sets custom environment variables.
+- `TEST_DOCKER_NO_PIN_PKGS` - If set, it will **not** pin the nixpkgs archive during testing.
+- `TEST_DOCKER_CUSTOM_PKGS` - Sets custom nix packages.
+- `TEST_DOCKER_EXPECTED_OUTPUT` - The expected output of the program that is being tested.
+- `TEST_DOCKER_WITH_POSTGRES` - If set, it will mount postgres to the test.
```
coffee-cup, line 67: Why not just use the same shared envs to configure nixpacks for the plan/build (e.g. envs, custom packages, etc)?

examples/node-nx-express/.eslintrc.json
```diff
@@ -0,0 +1,35 @@
+{
```
coffee-cup, line 1: Why were all the Nx files edited in this PR?

test-helper/src/lib.rs
devtomio: Its to remove code duplication, most tests are technically the same anyway, just different paths and outputs. Especially now with the `test.env` file is there, new contributors can add tests easily to the examples.

devtomio: If you want to, I can add another section to the Contributing Guide on how to add tests.

examples/node-nx-express/.eslintrc.json
devtomio: The NX tests (`node-nx-express`, `node-nx-next`, etc..) were all using the same path (`./examples/node-nx`), so I separated them up to be more verbose.

test-helper/src/lib.rs
coffee-cup: I don't think it removes code duplication that much though. The tests currently are only a few lines long
```ts
#[test]
fn test_node_custom_version() {
    let name = simple_build("./examples/node-custom-version");
    assert!(output.contains("Node version: v18"));
}
```

Moving that to a `test.env` file means that the test inputs are now separated across many different files. For the generate plan tests it makes sense because we just run a single function on all the examples. But the docker run tests have a lot more inputs and possibilities. Editing the docker container/run logic is less trivial without prior knowledge of how this works. It also means we cannot use the same example directory for multiple tests, which we do quite a bit.

coffee-cup:
Going to close this since even though it reduces some code duplication, it makes it significantly harder to reason about how the docker tests are running and where they come from. It also makes it harder to update how the tests are run in the future as everything is inside of a macro. 

Appreciate the work and effort you put into the PR though! In the future, for larger changes like this, It would be good to start with either an issue or discussion item to ensure that the change is something we want to make.

PR #425, opened 2022-08-19
Introduce Provider Metadata

AhmedMozaly:
Nixpacks can tell a lot about the codebase it's building. The programming language, the framework and more
But once the build is over and the container image is published, we lose these data and can't tell anymore what this container content is

Maintaining this data as part of the container image, can help us show better error messages if the container is not listing to the PORT and we know -for example- that it is Django app. we can't point the user to the right docs link that can help

This PR introduce new `ProviderMetadata` concept, where each provider adds the metadata it has detected. and Nixpacks will take it from here and combine all these metadata in one environment variable called `NIXPACKS_LABELS` for now


src/providers/mod.rs
```diff
@@ -24,4 +24,44 @@ pub trait Provider {
...
     fn detect(&self, app: &App, _env: &Environment) -> Result<bool>;
     fn get_build_plan(&self, _app: &App, _environment: &Environment) -> Result<Option<BuildPlan>>;
+    fn get_metadata(&self, _app: &App, _env: &Environment) -> Result<ProviderMetadata> {
+        Ok(ProviderMetadata::default())
+    }
+}
+
+#[derive(Default)]
+pub struct ProviderMetadata {
+    pub labels: Option<Vec<String>>,
```
Milo123459, line 34: Option<Vec is redundant, you should just use a Vec with nothing inside of it.

src/providers/mod.rs
```diff
@@ -24,4 +24,44 @@ pub trait Provider {
...
+        ProviderMetadata {
+            labels: Some(labels),
+        }
+    }
+
+    pub fn join_as_comma_separated(&self, provider_name: String) -> String {
+        let mut arr = vec![provider_name];
+        let mut labels_arr = match &self.labels {
+            Some(v) => v.clone(),
+            _ => vec![],
```
Milo123459, line 54:
```suggestion
            _ => Vec::new(),
```

src/providers/mod.rs
coffee-cup: `Option` is still useful if you ever want to deserialize data into a Rust struct since it will not error if the field is missing.

src/providers/mod.rs
```diff
@@ -24,4 +24,44 @@ pub trait Provider {
...
+        Ok(ProviderMetadata::default())
+    }
+}
+
+#[derive(Default)]
+pub struct ProviderMetadata {
+    pub labels: Option<Vec<String>>,
+}
+
+impl ProviderMetadata {
```
coffee-cup, line 37: Can you add some unit tests for this struct/functions

src/providers/python.rs
```diff
@@ -48,7 +60,8 @@ impl Provider for PythonProvider {
             plan.set_start_phase(start);
         }
 
-        if app.includes_file("poetry.lock") {
+        let metadata = self.get_metadata(app, env)?;
```
coffee-cup, line 63: Hmm I am unsure if it makes sense to use the metadata like this. We lose some type information here since metadata i just some key value pairs. Should all build plan decisions be made by first adding metadata? Where is the line drawn? Also, should the metadata be attached to the build plan? Or does it not affect the build plan at all? Currently it adds a environment variable with the metadata result, but maybe it is better to instead have a `nixpacks detect` or `nixpacks metadata` command that just returns the metadata as an object that we can use on Railway.

AhmedMozaly: Metadata should not be something "new", We already make decisions in build plans based on detected "facts"
This PR is just about moving the logic of detecting these "facts" to well defined place and give it a clear name. So we can collect from providers and make use of (like offering it in env var or use in Railway)

So, I would say that metadata should be part of build plan and even show in console output to let user know about what Nixpacks has detected! 

image

coffee-cup (approved):

PR #426, opened 2022-08-19
Java 8 -> 17

JakeCooper:
This PR lets us not use an ancient version 
of Java


coffee-cup (approved):
The snapshot tests need to be updated

PR #439, opened 2022-08-24
Read package lockfiles

JakeCooper:
This PR fixes a bug where we stopped parsing lockfiles to check dependencies


tests/snapshots/generate_plan_tests__node_canvas.snap
```diff
@@ -17,10 +17,10 @@ expression: plan
         "setup"
       ],
       "commands": [
-        "npm i"
+        "pnpm i --frozen-lockfile"
```
coffee-cup, line 20: The `node-canvas` example has all types of lockfiles. This isn't that realistic and it should probably just have a single one.

JakeCooper: Yea fair nuff

PR #440, opened 2022-08-25
Auto Publish Nixpacks using Labels

JakeCooper:
This PR should autopublish on merges to main

Initially done as a dry run then once validated will update

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,42 @@
+name: Tag
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions-ecosystem/action-get-merged-pull-request@v1
+        id: get-merged-pull-request
+        with:
+          github_token: ${{ secrets.GH_PAT }}
+
+      - uses: actions-ecosystem/action-release-label@v1
+        id: release-label
+        if: ${{ steps.get-merged-pull-request.outputs.title != null }}
+        with:
+          github_token: ${{ secrets.GH_PAT }}
+          labels: ${{ steps.get-merged-pull-request.outputs.labels }}
+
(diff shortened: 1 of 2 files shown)


Milo123459:
So whenever a PR is merged, a new release is made? Bad idea IMO

JakeCooper:
> So whenever a PR is merged, a new release is made? Bad idea IMO

Why?

Milo123459:
> Why?
Regressions could be introduced and a new release could be made with the regression. I think the best idea here is just some simple way of creating a release automatically, like some sort of commands to bump the version and make a new release automatically.

devtomio:
> > Why?
> 
> Regressions could be introduced and a new release could be made with the regression. I think the best idea here is just some simple way of creating a release automatically, like some sort of commands to bump the version and make a new release automatically.

+1, I think Nixpacks should have a release date like Railway (e.g. a changelog per week).

coffee-cup:
> > Why?
> 
> Regressions could be introduced and a new release could be made with the regression. I think the best idea here is just some simple way of creating a release automatically, like some sort of commands to bump the version and make a new release automatically.

We already have this currently with the `cargo release` command. It is just something that has to be run manually.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      
+      - run: echo 'Dry Run PATCH' # cargo release patch --execute
+        if: ${{ steps.release-label.outputs.level == release/patch }}
+      - run: echo 'Dry Run MINOR' # cargo release minor --execute
+        if: ${{ steps.release-label.outputs.level == release/minor }}
+      - run: echo 'Dry Run MAJOR' # cargo release major --execute
+        if: ${{ steps.release-label.outputs.level == release/major }}
```
coffee-cup, line 35: If you just leave out `--execute` it will do a dry run.

Also, does this trigger the release action? My understanding is that actions cannot trigger other actions.

PR #446, opened 2022-08-26
Add Easypanel 

deiucanta:
We've added Nixpacks in Easypanel. This PR adds Easypanel in the "Deploying" section of the docs.


coffee-cup (approved):
This is awesome! Exciting to see Nixpacks used on other platforms 😄

PR #448, opened 2022-08-27
Publish `.deb` package file in release

devtomio:
Publishing a `.deb` file in release allows Nixpacks to be installed with `dpkg`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -146,6 +146,7 @@ jobs:
       run: cargo deb --target x86_64-unknown-linux-musl --output nixpacks-${{ needs.create-release.outputs.nixpacks_version }}-amd64.deb
 
     - name: Upload .deb package file
+      if: matrix.target == 'x86_64-unknown-linux-musl'
       uses: svenstaro/upload-release-action@v2
       with:
         tag: ${{ needs.create-release.outputs.nixpacks_version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -137,6 +137,20 @@ jobs:
...
+    - name: Install cargo-deb
+      if: matrix.target == 'x86_64-unknown-linux-musl'
+      run: cargo install cargo-deb
+
+    - name: Generate .deb package file
+      if: matrix.target == 'x86_64-unknown-linux-musl'
+      run: cargo deb --target x86_64-unknown-linux-musl --output nixpacks-${{ needs.create-release.outputs.nixpacks_version }}-amd64.deb
+
+    - name: Upload .deb package file
+      uses: svenstaro/upload-release-action@v2
```
nebulatgs, line 149:
```suggestion
      if: matrix.target == 'x86_64-unknown-linux-musl'
      uses: svenstaro/upload-release-action@v2
```
dont we need to check here as well?

nebulatgs, line 149:
```suggestion
      if: matrix.target == 'x86_64-unknown-linux-musl'
      uses: svenstaro/upload-release-action@v2
```
dont we need to check here as well?

nebulatgs (changes requested):

.github/workflows/release.yml
nebulatgs: since otherwise the `.deb` file won't exist

devtomio: oops, thank you

coffee-cup (approved):

PR #453, opened 2022-08-27
Improve environment variable parameter parsing

### LINKED ISSUE #445: Unable to parse variable string

### Describe the bug

I'm trying to pass some weird env variables for a Laravel app but I get "Unable to parse variable string".

```
nixpacks build . --name testing --env "APP_KEY=base64:abcdef...="
```

It doesn't like the final `=` character.

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

-

Milo123459:
## What does this PR address?

Fixes #445

Improves parsing by using a Regex.

### DIFF
--- a/src/nixpacks/environment.rs
+++ b/src/nixpacks/environment.rs
@@ -1,4 +1,5 @@
 use anyhow::{bail, Result};
+use regex::Regex;
 use std::{collections::BTreeMap, env};
 
 pub type EnvironmentVariables = BTreeMap<String, String>;
@@ -16,19 +17,25 @@ impl Environment {
     pub fn from_envs(envs: Vec<&str>) -> Result<Environment> {
         let mut environment = Environment::default();
         for env in envs {
-            let v: Vec<&str> = env.split('=').collect();
-            if v.len() == 1 {
+            let matches = Regex::new(r"([^=]*)=([\s\S]*)")
+                .unwrap()
+                .captures(env)
+                .unwrap();
+            if matches.len() == 1 {
+                let name = matches.get(1).unwrap().as_str();
                 // Pull the variable from the current environment
-                let name = v[0];
                 if let Ok(value) = env::var(name) {
                     // Variable is set
                     environment.set_variable(name.to_string(), value);
                 }
-            } else if v.len() > 2 {
+            } else if matches.len() > 2 {
                 bail!("Unable to parse variable string");
             } else {
                 // Use provided name, value pair
-                environment.set_variable(v[0].to_string(), v[1].to_string());
+                environment.set_variable(
+                    matches.get(1).unwrap().as_str().to_string(),
+                    matches.get(2).unwrap().as_str().to_string(),
+                );
             }
         }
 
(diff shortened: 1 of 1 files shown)


nebulatgs:
be more descriptive in pull requests (at least summarize what its fixing)

src/nixpacks/environment.rs
```diff
@@ -16,19 +17,26 @@ impl Environment {
     pub fn from_envs(envs: Vec<&str>) -> Result<Environment> {
         let mut environment = Environment::default();
         for env in envs {
-            let v: Vec<&str> = env.split('=').collect();
-            if v.len() == 1 {
+            let matches = Regex::new(r"([^=]*)(?:=?)([^=]*)([\s\S]*)")
+                .unwrap()
+                .captures(env)
+                .unwrap();
```
coffee-cup, line 23: Should not `unwrap` here. Or anywhere below either.

src/nixpacks/environment.rs
```diff
@@ -16,19 +17,26 @@ impl Environment {
     pub fn from_envs(envs: Vec<&str>) -> Result<Environment> {
         let mut environment = Environment::default();
         for env in envs {
-            let v: Vec<&str> = env.split('=').collect();
-            if v.len() == 1 {
+            let matches = Regex::new(r"([^=]*)(?:=?)([^=]*)([\s\S]*)")
```
coffee-cup, line 20: I don't think this regex is correct. Environment variable names cannot have any character. We should limit it to alphanum characters

coffee-cup (commented):
Please add some tests for the new variable parsing

src/nixpacks/environment.rs
Milo123459: I was mirroring behaviour that was already there, which was splitting which would account for any character. But I'll implement.

src/nixpacks/environment.rs
Milo123459: In my tests, there can never be less then 1 match so it's fine to unwrap.

src/nixpacks/environment.rs
```diff
@@ -16,19 +17,26 @@ impl Environment {
     pub fn from_envs(envs: Vec<&str>) -> Result<Environment> {
         let mut environment = Environment::default();
         for env in envs {
-            let v: Vec<&str> = env.split('=').collect();
-            if v.len() == 1 {
+            let matches = Regex::new(r"([A-Za-z0-9_-]*)(?:=?)([A-Za-z0-9_-]*)([\s\S]*)")
```
nebulatgs, line 20:
```suggestion
            let matches = Regex::new(r"([A-Za-z0-9_-]*)(?:=?)([^=]*)([\s\S]*)")
```
the variable body can be anything that isnt an `=`

nebulatgs (dismissed):

src/nixpacks/environment.rs
```diff
@@ -86,11 +90,6 @@ mod tests {
         assert!(environment.get_variable("NON_EXISTANT").is_none());
     }
 
-    #[test]
-    fn test_create_invalid_environment() {
-        assert!(Environment::from_envs(vec!["INVALID=ENV=CONFIG"]).is_err());
-    }
-
```
coffee-cup, line 93: Instead of removing this test can you instead assert that it is valid and that the name/value are correct

Milo123459: done

coffee-cup (approved):

src/nixpacks/environment.rs
Milo123459: done

Issue #462, opened 2022-08-28
Support Pandas/Numpy


JakeCooper:
### Feature request

Various errors seem to occur with Pandas/Numpy, related to the following error

`ImportError: libstdc++.so.6: cannot open shared object file: No such file or directory`

Related Threads: 
- Discord: [here](https://discord.com/channels/713503345364697088/1013124535949733898/1013555831758917693)
- Twitter: [here](https://twitter.com/somil_agr/status/1563392838418976771)

It appears, going off this [nix forums post](https://discourse.nixos.org/t/python-package-with-runtime-dependencies/5522/7) we need to set the LD_LIBRARY. 

We already do this in Nixpacks ([here](https://github.com/railwayapp/nixpacks/blob/c51d381854dda521a45c2d0ddfca0f36f581d9a8/src/nixpacks/nix/mod.rs#L58)), probably need to make this more configurable or set-able on the provider level

If someone has a repo example that would be great, and would obviously love to take a PR for this

coffee-cup:
Hmm we do already have custom support for numpy. I guess needs to be updated. https://github.com/railwayapp/nixpacks/blob/main/src/providers/python.rs#L107

coffee-cup:
Perhaps the detection is slightly off

Entropy5:
Just tested a pandas repo again and it still crashes with this error 
https://github.com/Entropy5/plotly-test

coffee-cup:
I was just able to build and run that repo

<img width="1288" alt="image" src="https://user-images.githubusercontent.com/3044853/187341820-3ae026ff-2e59-4d9c-85c6-a158bb52b6e6.png">

What version of Nixpacks do you see an error with? And can you please share the full build logs (including build plan)

coffee-cup:
Should be fixed in the latest version of nixpacks

Entropy5:
I was using the railway browser implementation to deploy this straight from github with all default settings
I will try the commandline thing now

AndreyBoyarchuk:
[...]

PR #463, opened 2022-08-29
add zlib nix libraries if pandas is detected

coffee-cup:
Previously we were just adding the `zlib` library if `numpy` was detected. This PR changes it so we also add them if only `pandas` is detected.

Fixes https://github.com/railwayapp/nixpacks/issues/462


JakeCooper (approved):
This won't actually solve the linked issue though correct? Cause we need libstdc++ which requires LD_LIBRARY updates

coffee-cup:
> This won't actually solve the linked issue though correct? Cause we need libstdc++ which requires LD_LIBRARY updates

`LD_LIBRARY` is already updated with Nix libraries that the plan specifies. `zlib` is the Nix package needed.

Issue #373, opened 2022-07-29
Python install relative packages


coffee-cup:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When installing Python deps from the `requirements.txt` file we only copy over that file. However, this is a problem when installing relative dependencies. (https://canary.discord.com/channels/713503345364697088/872280248807469106/1002517924323545179)

We should copy over all files before installing so we can support this.

Milo123459:
Rather then copying over all files, we can just detect when the `file:` protocol is used and just copy over the file after the protocol.

astrojuanlu:
I confirm this fixed the issue 🎉 Thanks!

PR #464, opened 2022-08-29
Allow installing Python relative deps

coffee-cup:
This PR allows installing Python relative deps by adding all files to the container before installing. With the Python dep cached installing is very quick.

Fixes https://github.com/railwayapp/nixpacks/issues/373


astrojuanlu:
The best fixes are those that happen by _just_ removing code 😄

PR #468, opened 2022-08-30
Install gcc and clibs for python provider by default

coffee-cup:
Many (numpy, pandas, other ml libs) require gcc and c libraries to be available. We've tried matching on these packages in `requirements.txt` but it seems like a never ending game of cat and mouse. This PR installs `gcc` and some required Nix libraries by default in all Python containers. I'm hoping this reduces the number of random (missing C package errors for Python).

### DIFF
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -97,16 +97,16 @@ impl PythonProvider {
 
         if PythonProvider::is_django(app, env)? && PythonProvider::is_using_postgres(app, env)? {
             // Django with Postgres requires postgresql and gcc on top of the original python packages
-            pkgs.append(&mut vec![Pkg::new("postgresql"), Pkg::new("gcc")]);
+            pkgs.append(&mut vec![Pkg::new("postgresql")]);
         }
 
         let mut setup_phase = Phase::setup(Some(pkgs));
 
-        // Numpy needs some C headers to be available
+        // Many Python packages need some C headers to be available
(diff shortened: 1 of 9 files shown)


src/providers/python.rs
```diff
@@ -97,16 +97,16 @@ impl PythonProvider {
...
 
         let mut setup_phase = Phase::setup(Some(pkgs));
 
-        // Numpy needs some C headers to be available
+        // Many Python packages need some C headers to be available
         // stdenv.cc.cc.lib -> https://discourse.nixos.org/t/nixos-with-poetry-installed-pandas-libstdc-so-6-cannot-open-shared-object-file/8442/3
-        if PythonProvider::uses_dep(app, "numpy")? || PythonProvider::uses_dep(app, "pandas")? {
-            setup_phase.add_pkgs_libs(vec!["zlib".to_string(), "stdenv.cc.cc.lib".to_string()]);
-        }
+        println!("Numpy or Pandas detected, adding gcc and libstdc++");
```
JakeCooper, line 107: 🔫

JakeCooper (approved):

src/providers/python.rs
```diff
@@ -323,6 +322,7 @@ impl PythonProvider {
         ))
     }
 
+    #[allow(dead_code)]
```
Milo123459, line 325: Do we need this? Is there a time when this function is never called?

coffee-cup: It currently is never called, but I didn't want to remove it since it may be useful in the future.

PR #477, opened 2022-08-30
Support installing Ruby local deps

coffee-cup:
This PR copies over all files before running `bundle install` so that local deps can be installed


Milo123459:
Should we add a test for local-deps?

coffee-cup:
> Should we add a test for local-deps?

Done

PR #478, opened 2022-08-30
Delete tmp directory after successful build

AhmedMozaly:
Nixpacks creates a temp directory (if `--out` was not provided) then it takes a copy of the code & build at this temp directory 
This PR ensures that we remove this temp directory after we're done.


coffee-cup (approved):

JakeCooper:
There's no performance penalties with this right? Like we're not removing a cache or something it's literally just the stuff we artifact

PR #480, opened 2022-08-31
Add Elixir support to docs

AhmedMozaly:
This PR updates docs with Elixir support

### DIFF
--- a/docs/pages/docs/providers/elixir.md
+++ b/docs/pages/docs/providers/elixir.md
@@ -1,0 +1,28 @@
+---
+title: Elixir
+---
+
+# {% $markdoc.frontmatter.title %}
+
+Elixir is detected if a `mix.exs` file is found.
+
+## Setup
+```
+mix local.hex --force
+mix local.rebar --force
+mix deps.get --only prod
+```
+
+## Build
+
+```
+mix compile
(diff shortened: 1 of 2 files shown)


docs/pages/docs/providers/elixir.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Elixir
+---
+
+# {% $markdoc.frontmatter.title %}
+
+Elixir is detected if a `mix.exs` file is found.
```
coffee-cup, line 7: I think we should mention that Elixir with a DB is not currently supported

coffee-cup (approved):

docs/pages/docs/providers/elixir.md
JakeCooper: A note at the bottom would be good. That said, if someone comes with a broken Elixir with database we can fix it for them ASAP

Maybe a note about reaching out on Discord?

AhmedMozaly: I've just created a PR to add Ecto & Postgress support to Elixir
https://github.com/railwayapp/nixpacks/pull/485

Will merge current PR for now and can add this bottom line if we the other PR took longer to merge

PR #489, opened 2022-08-31
Add current-dir flag

AhmedMozaly:
This PR adds new `--current-dir` flag. When passed, Nixpacks won't create a temp dir & copy code to
Will start working right away on the current directory and create `.nixpacks` sub-directory at.

Also, with `--current-dir` flag. Nixpacks won't delete the directory after successful build, It will be up to the caller to decide.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -85,8 +85,7 @@ fn main() -> Result<()> {
                     Arg::new("current-dir")
                         .long("current-dir")
                         .help("Output Nixpacks related files to the current directory ")
-                        .takes_value(false)
-                        .global(true),
+                        .takes_value(false),
                 )
                 .arg(
                     Arg::new("no-cache")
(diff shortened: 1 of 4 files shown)


Milo123459:
Is there a real reason to use this?

AhmedMozaly:
Using Nixpacks on server side scenario (Like Railway), The code is already cloned to a target directory even before calling Nixpacks. 

Creating new temp directory & copying all files to, Would be redundant work.

src/main.rs
```diff
@@ -81,6 +81,13 @@ fn main() -> Result<()> {
                         )
                         .takes_value(true),
                 )
+                .arg(
+                    Arg::new("current-dir")
+                        .long("current-dir")
+                        .help("Output Nixpacks related files to the current directory ")
+                        .takes_value(false)
+                        .global(true),
```
coffee-cup, line 89: This flag doesn't need to be globally and should only be available for the `build` command

coffee-cup (approved):

Issue #488, opened 2022-08-31
Trouble Deploying Django App That Uses An External SQl Database


MCole-K9:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

build fails while trying to deploy my Django application. 

For context: I'm using planetscale for the mysql database and I use pipenv to create my virtual environment locally

Here is the build log:

``` 
#10 [stage-0 6/8] COPY . /app/
#10 sha256:45b179abd76949bbb9d7e4b3059bbac293597b647224e6037f1f5470615c9ec3
 
#10 DONE 0.1s
 
#11 [stage-0 7/8] RUN --mount=type=cache,id=s/7cba9214-d460-4f63-9dab-0761a89e5b3e-/root/cache/pip,target=/root/.cache/pip python -m venv /opt/venv && . /opt/venv/bin/activate && pip install -r requirements.txt
#11 sha256:f7747eb633f7186c74c5beea6c467701fbdad7ca6a1b416fb1e863f8af1e2620
 
#11 5.507 Collecting asgiref==3.5.2
 
#11 5.511   Using cached asgiref-3.5.2-py3-none-any.whl (22 kB)
 
#11 5.716 Collecting Django==4.1
#11 5.772   Using cached Django-4.1-py3-none-any.whl (8.1 MB)
 
#11 5.909 Collecting django-environ==0.9.0
 
#11 5.913   Using cached django_environ-0.9.0-py2.py3-none-any.whl (17 kB)
#11 5.949 Collecting django-psdb-engine==1.0.6
#11 5.952   Using cached django_psdb_engine-1.0.6-py2.py3-none-any.whl (3.7 kB)
#11 6.009 Collecting gunicorn==20.1.0
#11 6.014   Using cached gunicorn-20.1.0-py3-none-any.whl (79 kB)
 
#11 6.073 Collecting mysqlclient==2.1.1
#11 6.077   Using cached mysqlclient-2.1.1.tar.gz (88 kB)
#11 6.100   Preparing metadata (setup.py): started
 
#11 6.325   Preparing metadata (setup.py): finished with status 'error'
#11 6.337   error: subprocess-exited-with-error
#11 6.337
#11 6.337   × python setup.py egg_info did not run successfully.
#11 6.337   │ exit code: 1
#11 6.337   ╰─> [16 lines of output]
[...]

Issue #483, opened 2022-08-31
Support MySQL autodetect with Django/Python


JakeCooper:
### Feature request

Need to add auto detection of mysql client for Django

Related thread on Discord [here](https://discord.com/channels/713503345364697088/1014002182196043846/1014195641590751362)

### Contribution

If someone could give an example repo of this failing that would be a great start

MCole-K9:
Here is the  example repo - https://github.com/MCole-K9/Django-Railway. 

Project ID:
ec74eabd-8f3b-4e44-b405-8bfbcf517af6.

PR #496, opened 2022-09-02
python provider: support mysql backends for django

char8:
Fixes https://github.com/railwayapp/nixpacks/issues/483 and Fixes https://github.com/railwayapp/nixpacks/issues/488 by providing the dependency required by the `mysqlclient` python package.

Tested building the example linked in https://github.com/railwayapp/nixpacks/issues/488 (🙏 ) locally and running the resulting container against a test PlanetScale DB. Assuming it works the same for a generic MySQL/MariaDB server.


coffee-cup:
This is awesome! Can you please add a test + example app that uses Django and mysql. For tests, I think we should have
- Unit test in the Python provider to check that `is_using_mysql` works
- Generate plan test (generated automatically) to check that the build plan contains the correct package (https://github.com/railwayapp/nixpacks/blob/main/CONTRIBUTING.md#snapshot-tests)
- Docker run test to check that the container can be built and run if it uses MySQL (can see how the Postgres Django test works)

tests/docker_run_tests.rs
```diff
@@ -245,6 +247,63 @@ fn run_postgres() -> Container {
...
+        .spawn()
+        .unwrap()
+        .wait()
+        .context("starting mysql")
+        .unwrap();
+
+    // MySQL starts listening for connections after it has initialised its default database
+    // give it enough time to start before we continue with starting the application
+    // TODO: swap this for a healthceck + retry-loop because the arbitrary delay can lead to flakey tests
+    thread::sleep(Duration::new(10, 0));
```
char8, line 290: @coffee-cup thanks for the feedback - implemented the tests. But MySQL is being a bit annoying here, will address the TODO and ping you once it's ready for review.

Thinking of running a retry-loop with timeout (say 30s) of a command to check if MySQL is ready or not inside this function.

examples/python-django-mysql/mysite/settings.py
```diff
@@ -0,0 +1,127 @@
...
+
+WSGI_APPLICATION = 'mysite.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/4.0/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.mysql',
```
char8, line 78: Copy of the django-postgres example with this line and L4 of requirements.txt updated

tests/docker_run_tests.rs
char8: actually - looking at the runtimes, it seems like some of the docker tests take a few mins to complete. Could also add some margin to the sleep  (eg: make it 20s) and leave it as-is to reduce complexity if we're ok with that.

char8: whoops missed the unit test

coffee-cup (approved):
Awesome! 🎉

PR #499, opened 2022-09-05
Pass environment variables to PHP FPM

deiucanta:
PHP FPM doesn't pass env variables by default but I think it's helpful to disable this behavior. In Docker, a lot of times you want to pass variables via the `-e` flag.

Here is an article about this issue: https://medium.com/@tomahock/passing-system-environment-variables-to-php-fpm-when-using-nginx-a70045370fad


deiucanta:
@coffee-cup I'm not a Rust dev and I don't know how to fix the test snapshots. Can you do it?

aleksrutins:
Add `\nclear_env=no\n` to `staticAssets["php-fpm.conf"]` in the PHP test snapshots.

deiucanta:
@aleksrutins done

coffee-cup (approved):
Thanks 😄

PR #507, opened 2022-09-06
Error when no start command is found

Milo123459:
This PR errors when no start command can be found, as opposed to giving an un-descriptive error message. This should hopefully be clearer for the end user.

### DIFF
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use std::path::Path;
 use std::{env, fs, path::PathBuf};
 
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use globset::Glob;
 use regex::Regex;
 use serde::de::DeserializeOwned;
@@ -171,7 +171,10 @@ impl App {
 
     /// Convert an absolute path to a path relative to the app source directory
     pub fn strip_source_path(&self, abs_path: &Path) -> Result<PathBuf> {
-        let source_str = self.source.to_str().unwrap();
+        let source_str = match self.source.to_str() {
+            Some(s) => s,
+            None => bail!("Failed to parse source path"),
+        };
 
         // Strip source path from absolute path
         let stripped = match abs_path.strip_prefix(source_str) {
--- a/src/nixpacks/app.rs
+++ b/src/nixpacks/app.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use std::path::Path;
 use std::{env, fs, path::PathBuf};
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use globset::Glob;
 use regex::Regex;
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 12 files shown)


src/nixpacks/builder/docker/dockerfile_generation.rs
```diff
@@ -232,7 +232,7 @@ impl DockerfileGenerator for StartPhase {
     ) -> Result<String> {
         let start_cmd = match &self.cmd {
             Some(cmd) => utils::get_exec_command(cmd),
-            None => "".to_string(),
+            None => bail!("No start command cound be found"),
```
coffee-cup, line 235: The dockerfile generation should not error. It should error when the build plan is created/validated

src/nixpacks/app.rs
```diff
@@ -171,10 +171,7 @@ impl App {
 
     /// Convert an absolute path to a path relative to the app source directory
     pub fn strip_source_path(&self, abs_path: &Path) -> Result<PathBuf> {
-        let source_str = match self.source.to_str() {
-            Some(s) => s,
-            None => bail!("Failed to parse source path"),
-        };
+        let source_str = self.source.to_str().unwrap_or("");
```
coffee-cup, line 174: Why default to empty string here?

Milo123459: Oops, got left in.

Milo123459: Fixed in [`c236391` (#507)](https://github.com/railwayapp/nixpacks/pull/507/commits/c2363914f6b5e09ed8ad050499fd5a29d492a59b)

src/nixpacks/app.rs
```diff
@@ -171,10 +171,7 @@ impl App {
 
     /// Convert an absolute path to a path relative to the app source directory
     pub fn strip_source_path(&self, abs_path: &Path) -> Result<PathBuf> {
-        let source_str = match self.source.to_str() {
-            Some(s) => s,
-            None => bail!("Failed to parse source path"),
-        };
+        let source_str = self.source.to_str().unwrap();
```
coffee-cup, line 174: Why change it from what it was before? Which was a `bail!` with error message

coffee-cup: Unwrapping will also panic as opposed to returning a `Result`

coffee-cup (commented):
What does it look like if there is no start command found? I think we might want to
1. Only error if building and there is no start command (not while running the plan command). Happy to hear opposing thoughts though
2. Show the build plan in the current state if there is no start command found

src/nixpacks/app.rs
Milo123459: Resolved

coffee-cup:
It actually might make sense to move the start command check here https://github.com/railwayapp/nixpacks/blob/main/src/lib.rs#L95

and then also print the build plan before bailing

Milo123459:
What is the benefit of moving it there?

coffee-cup:
> What is the benefit of moving it there?

It will only error when actually building and won't error when running `nixpacks plan`

coffee-cup:
It is also not really the responsibility of the build plan generator to check if a start command exists.

Milo123459:
Done, moved logic.

coffee-cup (approved):

Issue #455, opened 2022-08-27
Support reading the Java version from `system.properties` file


coffee-cup:
The `system.properties` file in Java projects can contain `java.runtime.version=11` which indicates the Java version to use. We should support reading that and setting the java version based on the version defined there.

https://howtodoinjava.com/java/basics/java-system-properties/

Sodawyx:
Has the issue been solved? It seems that Nixpacks cannot recognize the java version from `system.properties` file automatically.

PR #511, opened 2022-09-07
Enhance Gradle provider

AhmedMozaly:
Always installing the latest version of `jdk` could break old Gradle apps, [Certain Gradle versions needs certain versions of `jdk`](https://docs.gradle.org/current/userguide/compatibility.html)

This PRs do:
* Try to read the Gradle version from `gradle/wrapper/gradle-wrapper.properties` and choose the right `jdk` version for
* Ensure Gradle caching is working as expected by targeting `~/.gradle` 
* Detect Spring Boot and pass `-Dserver.port=$PORT` to
* Update the `java.md` with these recent changes


src/providers/java.rs
```diff
@@ -29,15 +31,17 @@ impl Provider for JavaProvider {
...
+        let mut setup: Phase;
         let build = if self.is_using_gradle(app) {
-            setup.add_nix_pkgs(vec![Pkg::new("gradle")]);
+            let pkgs = self.get_jdk_and_gradle_pkgs(app)?;
+            setup = Phase::setup(Some(pkgs));
+
             let gradle_exe = self.get_gradle_exe(app);
             let mut build = Phase::build(Some(format!("{} build", gradle_exe)));
-            build.add_cache_directory(".gradle");
+            build.add_cache_directory("~/.gradle");
```
coffee-cup, line 41: Did you confirm this works? I'm not sure the `~` works in cache locations.

src/providers/java.rs
```diff
@@ -29,15 +31,17 @@ impl Provider for JavaProvider {
...
             build
         } else {
+            setup = Phase::setup(Some(vec![Pkg::new("jdk")]));
```
coffee-cup, line 44: Unrelated to gradle, but we might also want to check for a specific version of the `jdk` (https://github.com/railwayapp/nixpacks/issues/455)

I'm only mentioning since you are in Java land already :)

coffee-cup (approved):

src/providers/java.rs
AhmedMozaly: here https://github.com/railwayapp/nixpacks/blob/main/src/nixpacks/builder/docker/utils.rs#L11

the `~` is being replaced with `root`
Will push now a more clear path to avoid this confusion 


This screenshot is from inside the container, as you can see the cache exists at `/root/.gradle`
image

src/providers/java.rs
AhmedMozaly: Awesome!
I'm still in the Java context, let me shoot another PR for this to decrease the noise here!

src/providers/java.rs
AhmedMozaly: Ok, seems that we have both `/root/.gradle` & `.gradle` in the container 
After more investigation, the active one is `.gradle` I've changed it back now

Issue #494, opened 2022-09-02
Allow specifying nixpacks config in Cargo.toml package.metadata section


lilyball:
### Feature request

When building a Rust project, it would be really handy if I could add a `[package.metadata.nixpacks]` section to my `Cargo.toml` and use that to specify things like libs, apt, pkgs, etc. This way the project specifies its dependencies and can be built directly with `nixpacks` without having to pass separate env vars or flags.

This is a Rust-specific solution, and maybe it's better to pursue a language-independent one such as reading an `.env` file for the `NIXPACKS_*` env vars, or letting me define a `nixpacks.nix` file that's a nix module that provides definitions for config keys (e.g. a file containing `{ pkgs, ... }: { pkgs = [ pkgs.jq ]; }`), but the Rust solution is pretty straightforward and could be reproduced in any language that supports structured metadata in package info.

### Motivation

This would really help right now because I'm trying to use `flyctl deploy --nixpacks` and I can't find any way to pass nixpacks configuration via `flyctl`. That is undoubtedly something that `flyctl` should support, but a solution on the nixpacks side would make it easier for wrapper tools (like `flyctl`) to work.

### Contribution

Sorry

JakeCooper:
We're chatting a bit about this internally, and it also came up in the CLI repo as well

https://github.com/railwayapp/cli/issues/263

We'd love to work on an actual IaC representation. We've fiddled with it internally, but I'd like to start tracking requirements/demand externally as well

https://feedback.railway.app/feature-requests/p/infrastructure-as-code

IgnisDa:
I suggest we introduce a `nixpacks.toml` which can allow to allow these configurations.

coffee-cup:
> I suggest we introduce a `nixpacks.toml` which can allow these configurations.

I have an RFC coming for this soon 🙂

PR #513, opened 2022-09-08
Experimental file-based configuration

coffee-cup:
This PR implements the beginnings of https://github.com/railwayapp/nixpacks/discussions/508. The main features are

- File-based configuration with a `nixpacks.toml` or `nixpacks.json` file
- Consistent configuration via the CLI flags, environment variables, and file
- Build plan merging
- Pin to nixpacks archive for all Nix installs
- Phases are stored as an object on the build plan
- Procfile support is implemented as a provider
- `nixpacks plan . -f toml` to output the plan in toml format

When using file-based config, a warning will appear mentioning that it is experimental and may change. It is marked as experimental so that we can have breaking changes and not be locked into a format just yet.

image

An example configuration file is

```toml
[phases.setup]
# Extend what Nix packages get installed
nixPkgs = ["...", "cowsay"]

[phases.build]
# Override the build command
cmds = ["yarn run client:build"]

[start]
cmd = "yarn run client:start"
```

With this change, a `providers` field has been added to the build plan. Zero or one provider may be used at the moment. This allows you to specify that
No automatic detection or build plan should be used. Only the config from the `nixpacks.toml` file will be use
```toml
providers = []
```

Specify exactly what provider should be used
```toml
providers = ["python"]
```

If `providers` is `None`, then the default automatic detection will be used.

Fixes https://github.com/railwayapp/nixpacks/issues/494


AhmedMozaly (approved):
Nice! 🎉

PR #531, opened 2022-09-10
Ignore gitignored files

### LINKED ISSUE #457: Large build contexts due to not respecting `.gitignore`

### Describe the bug

Trying to build a local directory with large folders like NodeJS `node_modules` or Rust `target` will take a long time, since Nixpacks will attempt to copy those folders into a tempdir and into the build container.

### To reproduce

Create a new Rust project
Run `cargo build` once locally
Run `nixpacks build .` and see how it takes a long time

### Expected behavior

Nixpacks would respect the `.gitignore` file and only copy the files that are needed.

### Environment

Windows 11
Nixpacks v0.3.8

Milo123459:
This PR ignores gitignored files. Fixes #457


src/nixpacks/app.rs
```diff
@@ -74,11 +76,13 @@ impl App {
             None => return Ok(Vec::new()),
         };
 
-        let walker = WalkDir::new(&self.source);
+        let walker = WalkBuilder::new(&self.source)
+            .hidden(false)
```
coffee-cup, line 80: This will not copy over hidden files right? I don't think that is the behaviour we want.

src/nixpacks/files.rs
```diff
@@ -1,30 +1,34 @@
 use anyhow::Result;
+use ignore::WalkBuilder;
 use std::{fs, io, path::Path};
-use walkdir::WalkDir;
 
 pub fn recursive_copy_dir<T: AsRef<Path>, Q: AsRef<Path>>(source: T, dest: Q) -> Result<()> {
-    let walker = WalkDir::new(&source).follow_links(false);
+    let walker = WalkBuilder::new(&source)
+        .follow_links(false)
+        .hidden(false)
```
coffee-cup, line 8: Same as here

src/nixpacks/app.rs
Milo123459: I think this actually disables ignoring hidden files. It's a stupid way of naming things basically.

coffee-cup: Well that is confusing af 
image

Want to add a comment so it doesn't accidentally get removed in the future

coffee-cup (approved):

PR #552, opened 2022-09-14
fix: allow targets without a options attribute

FinnDore:
This pr makes it so the `options` attribute does not need to be present on a  `target` within a `project.json` ( NX )

This originally came from [here](https://discord.com/channels/713503345364697088/1009464427185193070/1017567215140216865) where Jaka had a valid project.json without a options attribute in his build target, adding an empty one like `"options": {}` satisfied NIXPACKS.

[I seem to be unable to run the NX tests on my branch neither main](https://app.warp.dev/block/ECEJtCOWwXpktlCvnOZO01). So ill look into that tomorrow afternoon if i cant find the cause asap ( Sleep is vital :))


FinnDore:
Need to swap out the builder for something that does not require a options attribute -_-

FinnDore:
Just pushed an astro project without the options attribute. Please not ive used a dumby start command as astro build a static site with this nx extension and as we dont care about the start command in this test its fine.

thanks to whoever has run ci. quick reaction times :)

FinnDore:
Interesting ive broken some of the other NX tests. ill try get the working on my mac, its an m1 so i am having issues with that atm

FinnDore:
Should be fixed. Caused by sleep deprivation 🙄

FinnDore:
Looks like its happy. Just talking to neb about some unwraps i noticed. Happy to fix them in this PR or create another just dont want to be spammy with PR's  https://discord.com/channels/713503345364697088/960605685920452669/1020105591210922055

src/providers/node/nx.rs
```diff
@@ -23,7 +23,7 @@ pub struct Targets {
 #[derive(Debug, Serialize, PartialEq, Eq, Deserialize)]
 pub struct Target {
     pub executor: String,
-    pub options: Options,
+    pub options: Option<Options>,
```
Milo123459, line 26: Could we rename the `Options` struct to be something like `NxDevOptions`?

Milo123459 (changes requested):

FinnDore:
Good idea. renamed!
Also spoke to neb and he suggested that a new pr should be made to fix the unwraps in the node provider.

Milo123459 (approved):

PR #557, opened 2022-09-15
Turborepo support

### LINKED ISSUE #456: First class support for Turborepo
### Feature request

[Turborepo](https://turborepo.org/) app support similar to how NX is supported with Nixpacks.

We should have a `NIXPACKS_TURBO_APP_NAME` configure variable that can be used to custom the build and start commands that are used.

aleksrutins:
closes #456

This PR adds Turborepo support to Nixpacks.


examples/fsharp-api/obj/fsac.cache
```diff
@@ -0,0 +1,2 @@
+17/06/2022 15:19:45
```
nebulatgs, line 1: why is this here

examples/fsharp-cli/obj/fsac.cache
```diff
@@ -0,0 +1,2 @@
+17/06/2022 15:19:45
```
nebulatgs, line 1: why is this here

nebulatgs (changes requested):

docs/pages/docs/providers/node.md
```diff
@@ -42,6 +43,8 @@ All dependencies found in `packages.json` are installed with either NPM, Yarn, o
 
 The build script found in `package.json` if it exists or if its an NX Monorepo `(npm|pnpm|yarn|bun) run build <NxAppName> --configuration=production`.
 
+Or, if it's a Turborepo monorepo (detected if `turbo.json` exists), the `build` pipeline will be called, if it exists; otherwise, the `build` script of the package referenced by `NIXPACKS_TURBO_APP_NAME` will be called, if the variable is provided; otherwise, it will fall back to the build script found in `package.json`.
```
Milo123459, line 46:
```suggestion
Or, if it's a Turborepo monorepo (detected if `turbo.json` exists), the `build` pipeline will be called (if it exists). Otherwise, the `build` script of the `package.json` referenced by `NIXPACKS_TURBO_APP_NAME` will be called, if `NIXPACKS_TURBO_APP_NAME` is provided and valid. Otherwise, it will fall back to the build script found in `package.json` at the monorepos root.
```

docs/pages/docs/providers/node.md
```diff
@@ -51,6 +54,10 @@ The start command priority is
   - If the app is a NextJS project: `npm run start`
   - If `targets.build.options.main` exists in the apps `Project.json`: `node <outputPath>/<mainFileName>.js` (e.g `node dist/apps/my-app/main.js`)
   - Fallback: `node <outputPath>/index.js` (e.g `node dist/apps/my-app/index.js`)
+- If it's a Turborepo
```
Milo123459, line 57:
```suggestion
- If Turborepo is detected
```

src/providers/node/turborepo.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::nixpacks::{app::App, environment::Environment};
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct TurboJson {
+    pub pipeline: HashMap<String, Pipeline>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+#[allow(non_snake_case)]
```
Milo123459, line 14:
```suggestion
```

src/providers/node/turborepo.rs
```diff
@@ -0,0 +1,49 @@
...
+pub struct Pipeline {
+    pub dependsOn: Option<Vec<String>>,
```
Milo123459, line 16:
```suggestion
    #[serde(rename = "dependsOn")]
    pub depends_on: Option<Vec<String>>,
```

src/providers/node/mod.rs
```diff
@@ -179,6 +192,18 @@ impl NodeProvider {
...
         }
 
+        if Turborepo::is_turborepo(app) {
+            let turbo_cfg = Turborepo::get_config(app)?;
+            let app_name = Turborepo::get_app_name(env);
+
+            if let Some(start_pipeline) = Turborepo::get_start_cmd(&turbo_cfg) {
+                return Ok(Some(start_pipeline));
+            } else if let Some(name) = app_name {
+                return Ok(Some(format!("npx turbo run {}:start", name)));
```
Milo123459, line 202: We should check for which package manager to use

src/providers/node/mod.rs
```diff
@@ -179,6 +192,18 @@ impl NodeProvider {
...
+            }
+            return Ok(Some("npx turbo run start".to_string()));
```
Milo123459, line 204: Same here

Milo123459 (changes requested):
I've added some changes to `dependsOn`, making it snake case, so if you apply one, the other suggestions need to be applied. Good work so far!

src/providers/node/mod.rs
```diff
@@ -99,6 +100,18 @@ impl Provider for NodeProvider {
         let mut build = if NodeProvider::is_nx_monorepo(app) {
             let app_name = NodeProvider::get_nx_app_name(app, env)?.unwrap();
             Phase::build(Some(format!("npx nx run {}:build:production", app_name)))
+        } else if Turborepo::is_turborepo(app) {
+            let turbo_cfg = Turborepo::get_config(app)?;
+            if let Some(build_cmd) = Turborepo::get_build_cmd(&turbo_cfg) {
+                Phase::build(Some(build_cmd))
+            } else if let Some(app_name) = Turborepo::get_app_name(env) {
+                Phase::build(Some(format!("npx turbo run {}:build", app_name)))
```
coffee-cup, line 108: Should this check to make sure that the `app_name` does have a build script?

src/providers/node/mod.rs
coffee-cup: +1. There are a few places it should use the package manager.

src/providers/node/mod.rs
```diff
@@ -96,15 +97,32 @@ impl Provider for NodeProvider {
...
+            } else {
+                Phase::build(None)
+            }
         } else if NodeProvider::has_script(app, "build")? {
             let pkg_manager = NodeProvider::get_package_manager(app);
             Phase::build(Some(format!("{} run build", pkg_manager)))
         } else {
             Phase::build(None)
         };
+        drop(dlx);
```
coffee-cup, line 125: I don't think this is needed.

src/providers/node/mod.rs
```diff
@@ -160,23 +180,49 @@ impl NodeProvider {
...
             if project_json.targets.build.executor == "@nrwl/next:build" {
-                return Ok(Some(format!("cd {} && npm run start", output_path)));
+                return Ok(Some(format!(
+                    "cd {} && {} run start",
+                    pkg_manager, output_path
+                )));
             }
 
             let main = project_json.targets.build.options.main;
+            let executor = if pkg_manager == "bun".to_string() {
```
coffee-cup, line 199: nit: Use the `executor` everywhere else that `node` is used for the start command (not just for NX apps).

src/providers/node/turborepo.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct TurboJson {
+    pub pipeline: HashMap<String, Value>,
+}
+
+pub struct Turborepo;
+
+impl Turborepo {
+    pub fn is_turborepo(app: &App) -> bool {
```
coffee-cup, line 17: Thoughts on having a way to skip turbo detection? Maybe something like `NIXPACKS_SKIP_TURBO`

Milo123459: Turborepo configuration files are pretty unique.. useless IMO.

docs/pages/docs/providers/node.md
aleksrutins: It doesn't actually check the validity of `NIXPACKS_TURBO_APP_NAME`. Should it? I don't think that the NX code does.

Milo123459:
Could you fix conflicts please?

docs/pages/docs/providers/node.md
Milo123459: It should. I'll implement support for Nx and validating the name in another PR.

productdevbook:
new version 1.5.1 -> https://turborepo.org/docs/handbook/deploying-with-docker#the-solution prune support

Milo123459:
@aleksrutins any updates?

src/providers/node/mod.rs
```diff
@@ -151,32 +169,56 @@ impl NodeProvider {
...
+                        package_json.workspaces.unwrap_or_default()
+                    },
+                    &name,
+                )? {
+                    return Ok(Some(format!(
+                        "{} --workspace {} run start",
+                        pkg_manager, name
+                    )));
+                }
+                println!("warning: app `{}` not found", name);
```
coffee-cup, line 201: I'm unsure about this type of warning. But I think it is okay for now

```suggestion
                println!("Warning: Turborepo app `{}` not found", name);
```

src/providers/node/turborepo.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn get_start_cmd(cfg: &TurboJson) -> Option<String> {
+        Turborepo::get_pipeline_cmd(cfg, "start")
+    }
+
+    pub fn get_app_name(env: &Environment) -> Option<String> {
+        env.get_config_variable("TURBO_APP_NAME")
+    }
+
+    pub fn has_app(app: &App, workspaces: Vec<String>, name: &str) -> Result<bool> {
+        //TODO: parallelize?
```
coffee-cup, line 57: The number of workspaces will be very small so I think it is okay to keep as is

src/providers/node/mod.rs
```diff
@@ -142,6 +145,21 @@ impl NodeProvider {
...
+            return if let Some(build_cmd) = Turborepo::get_build_cmd(&turbo_cfg) {
+                Ok(Some(build_cmd))
+            } else if let Some(app_name) = Turborepo::get_app_name(env) {
+                Ok(Some(format!("{} turbo run {}:build", dlx, app_name)))
+            } else if NodeProvider::has_script(app, "build")? {
+                let pkg_manager = NodeProvider::get_package_manager(app);
+                Ok(Some(format!("{} run build", pkg_manager)))
+            } else {
+                Ok(None)
+            };
```
coffee-cup, line 160: Can we default to the regular node build command behaviour below at this point? The logic is the same right now, but ideally we don't have to maintain two versions. So if there is no Turbo build pipeline and no build script in the Turbo app name, we fallback to regular Node provider build command.

src/providers/node/mod.rs
```diff
@@ -151,32 +169,56 @@ impl NodeProvider {
...
+        let pkg_manager = NodeProvider::get_package_manager(app);
+        let executor = if pkg_manager == *"bun" { "bun" } else { "node" };
+        let package_json: PackageJson = app.read_json("package.json").unwrap_or_default();
+
         if Nx::is_nx_monorepo(app, env) {
             if let Some(nx_start_cmd) = Nx::get_nx_start_cmd(app, env)? {
                 return Ok(Some(nx_start_cmd));
             }
         }
+        if Turborepo::is_turborepo(app) {
```
coffee-cup, line 182: Can this block (and the build command block above), be moved to the `turborepo.rs` file (similar to how Nx works). 

```rs
if Turborepo::is_turborepo(app) {
  if let Some(turbo_start_cmd) = Turborepo::get_turbo_start_cmd(app, env, pkg_manager)? {
    return Ok(Some(turbo_start_cmd))
  }
}
```

It would also allow us to easily fallback to the regular node provider behaviour if no Turbo specific start/build commands can be found.

src/providers/node/mod.rs
```diff
@@ -151,32 +169,56 @@ impl NodeProvider {
...
+                        "{} --workspace {} run start",
+                        pkg_manager, name
+                    )));
+                }
+                println!("warning: app `{}` not found", name);
+            }
+            if let Some(start_pipeline) = Turborepo::get_start_cmd(&turbo_cfg) {
+                return Ok(Some(start_pipeline));
+            }
+            return Ok(Some(format!("{} turbo run start", dlx)));
```
coffee-cup, line 206: Should this check if the start pipeline exists?

tests/generate_plan_tests.rs
```diff
@@ -125,3 +125,14 @@ fn test_zig_gyro() {
         .unwrap()
         .contains("mkdir /gyro"));
 }
+
+#[test]
+fn test_node_turborepo_custom_app() -> Result<(), Box<dyn std::error::Error>> {
```
coffee-cup, line 130: Can just unwrap here to keep consistency with the other docker run tests

coffee-cup (commented):
This is looking good! Just a few comments/questions. I am mainly concerned with breaking existing Turbo apps that are being built with Nixpacks currently. In all cases that we can't find a Turbo specific command, we should fallback to the default Node provider behaviour (for the most part this PR does that, just had a few questions in some places).

src/providers/node/mod.rs
coffee-cup: Much cleaner! 🎉

coffee-cup (approved):
Awesome! Thanks again for adding this 😄

wladiston:
Any docs for this?

aleksrutins:
> Any docs for this?

See: https://nixpacks.com/docs/providers/node

wladiston:
Has it been tested with PNPM? I'm getting a weird error during the build phase

Screenshot 2022-12-01 at 9 55 00 am

AlfreddGco:
It seems like this is not working for pnpm (at least for me). After looking a bit into the code and the output of nixpack it would seem like it prioritizes pnpm over turbo and outputs pnpm command for build instead of turbo command, and hence, it doesn't find the "start" command of the target app in my monorepo

PR #581, opened 2022-09-24
feat: add support for COBOL

FinnDore:
# Add support for COBOL

Ive added a basic COBOL provider and am looking for some feedback as its very minimal RN.

## Some things for discussion:

- Would support for https://cobolget.com/ be good?, or ironically that might fall under the remit of the nodeProvider as COBOLGet is a CLI published to `npm` for COBOL. ( although its installation guide does not suggest to install COBOLGet locally with a `package.json`, so maybe its not a node provider thing ).
- I believe each COBOL is a separate thing ( you cant import COBOLS into other COBOLS ). Ive done a bit of looking and have seen some people concatenate many COBOLS into one output binary. We should support this right? [example](https://github.com/azac/cobol-on-wheelchair/blob/master/downhill.sh). Behind a flag or automatically if we find bultiple COBOLS.

Ive reached out to a COBOL community asking for their standard workflows, still awaiting a response, but the community is not that active so i don't hold much hope. And we have @MarkMeridius writes COBOL at work.

Gonna do some more Github trawling of COBOL codebases to find more use-cases this weekend / week.

[x] Tests are added/updated if needed 
[x] Docs are updated if needed


coffee-cup (commented):
I have never used Cobol before so can't comment much to the best practices in building and deploying Cobol apps to production (although maybe I'll start learning it after this gets merged 😄). If someone with more experience with Cobol wants to take a look, I'd appreciate it, but if not we can get this merged and iterate on it!

Could you also please update the snapshot tests. (https://github.com/railwayapp/nixpacks/blob/main/CONTRIBUTING.md#snapshot-tests

src/providers/cobol.rs
```diff
@@ -0,0 +1,83 @@
...
+        environment: &Environment,
+    ) -> anyhow::Result<Option<crate::nixpacks::plan::BuildPlan>> {
+        let setup = Phase::setup(Some(vec![Pkg::new("gnu-cobol"), Pkg::new("gcc")]));
+
+        let compile_args = environment
+            .get_config_variable(COBOL_COMPILE_ARGS)
+            .unwrap_or_else(|| DEFAULT_COBOL_COMPILE_ARGS.to_string());
+
+        let app_path = CobolProvider::get_app_path(self, app, environment);
+        let file_name = app_path.file_stem().unwrap().to_str().unwrap();
```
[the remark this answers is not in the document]
FinnDore: Good shout. ive defaulted the filename to an empty string. I dont know if i should also be printing a warning to the console. Im avoiding throwing and error here, but should i be?

FinnDore:
Me neither lol, and im getting the impression that COBOL varies alot between companies, for example it sounds like *cusotm cobol* compilers are somewhat common after speaking to Markey. Not even mentioning the proprietary ones. So ive decided the best starting point was GnuCOBOL.

But yea sounds like a good idea to get this merged then iterate on it in some way or another.

FinnDore:
windows path issues i think  😭

FinnDore:
Cant help noticing that `get_app_path` is looking like a bunch of triangles. Is there a preferred method for handling `options `and `results`? I guess its not to bad because there is only one path for the code to follow. Even clipy was pushing for the  `if let` syntax.

src/providers/cobol.rs
```diff
@@ -0,0 +1,115 @@
...
+            .unwrap_or_else(|| DEFAULT_COBOL_COMPILE_ARGS.to_string());
+
+        let app_path = CobolProvider::get_app_path(self, app, environment);
+        let mut file_name = "";
+
+        if let Some(file_stem) = app_path.file_stem() {
+            if let Some(unwrapped_file_name) = file_stem.to_str() {
+                file_name = unwrapped_file_name;
+            }
+        }
```
coffee-cup, line 49: Nit: This can be cleaned up a bit by using `and_then` and `unwrap_or_default`. Something like

```rs
 let file_name = app_path
  .file_stem()
  .and_then(|stem| stem.to_str())
  .unwrap_or_default();
```

A few of the other _triangles_ can also be de-trianglified using a similar approach

FinnDore: perfect! Ive removed the triangles using this method, and refactored the code a tad.

coffee-cup (approved):

coffee-cup:
Awesome! 🎉

PR #584, opened 2022-09-26
Add `release/patch` on update-nixpkgs-archive workflow

Milo123459:
This PR adds the `release/patch` label when creating PRs for updating the nixpkgs version.


coffee-cup (changes requested):
This should create a patch release since it actually affects the containers that are built

coffee-cup (approved):

PR #597, opened 2022-10-01
Python psycopg2 and postgres

coffee-cup:
This PR installs the `postgresql` Nix package if the python `psycopg2` dependency is detected

### DIFF
--- a/src/providers/python.rs
+++ b/src/providers/python.rs
@@ -112,8 +112,8 @@ impl PythonProvider {
 
         pkgs.append(&mut vec![python_base_package]);
 
-        if PythonProvider::is_django(app, env)? && PythonProvider::is_using_postgres(app, env)? {
-            // Django with Postgres requires postgresql and gcc on top of the original python packages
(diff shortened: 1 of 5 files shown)


src/providers/python.rs
```diff
@@ -220,7 +220,9 @@ impl PythonProvider {
         // Check for the engine database type in settings.py
         let re = Regex::new(r"django.db.backends.postgresql").unwrap();
 
-        app.find_match(&re, "/**/*.py")
+        let uses_pg =
+            app.find_match(&re, "/**/*.py")? || PythonProvider::uses_dep(app, "psycopg2")?;
```
Milo123459, line 224: This should use a check dep function imo.

Milo123459 (changes requested):

src/providers/python.rs
coffee-cup: It does. `PythonProvider::uses_dep`. The `find_match` is looking for a specific file name

Milo123459 (approved):

PR #599, opened 2022-10-01
use absolute paths for compatibility with podman

Lunarequest:
This PR... resolves railwayapp/nixpacks/discussions/121 by using absolutes paths over relative paths for mount points as podman is currently incompatible with docker's behavior. There is an issue upstream to fix this but it would be faster to work around the issue downstream rather then fix it upstream in buildah and waiting for a podman issue to include it.


Milo123459 (approved):

Milo123459:
Could it potentially be worth adding Podman tests in CI? cc @coffee-cup

Lunarequest:
In my opinion it would not be worth it for the following reasons.
1. Podman usually is 100% compatible with docker
2. This incompatibility was the result of undocumented behavior on dockers side
3. Other then this the only compatibility issue would be new features added to docker which can't be fixed down stream.

Milo123459:
If you can figure out a way to make all the docker tests run with podman as well (and in parallel, so the docker and podman tests run at the same time) then I think that would be a good idea.

Lunarequest:
I will do my best, is there an irc/matrix/discord where I could ask questions

Milo123459:
https://discord.gg/railway

Milo123459:
Yeah, the account pushing the commit isn't on GitHub. I will trigger it.

Lunarequest:
not at all :/

PR #638, opened 2022-10-28
Experimental Support for Multiple Providers

coffee-cup:
This PR adds experimental support for generating a build plan that uses multiple providers. The providers must explicitly be set in the `nixpacks.toml` file. Extending the auto detected provider is supported with the `"..."` syntax.

### Phase names

The phases from the first provider in the list will keep the same names that they were created with. Subsequent providers phases will be prefixed with the name of the provider (e.g. `rust:setup`). This is done since phase names cannot have collisions as well as to correctly sort the phase build order. The rust build phase does not depend on the node install phase, and prefixing the phase names ensures that dependency isn't created.

One downside of this approach is that to override the behaviour of a specific non-first phase, you have to do something like
```toml
[phases."rust:build"]
cmds = ["..."]
```

which is a bit awkward. This is only because of allowed characters in the toml language. If the config format is Json, then it is a bit cleaner.

Suggestions of another prefix character besides `:` is welcome.

### Examples

This plan will use **only** the node and rust providers

```toml
providers = ["node", "rust"]
```

This plan will use the auto detected provider and python

```toml
providers = ["...", "python"]
```

### Why is this experimental

While testing this, I've noticed that many of the providers do not handle when specific language files (e.g. package.json) are not found when creating the build plan. Install, build, and start phases should only be added to the plan when appropriate. Some providers assume that since it is running, then there will always be an install, build, or start phases. These kinks will be discovered through use of multiple providers in different app scenarios. When using multiple providers, there is a warning that there might be some bugs.

image


gschier (approved):
Super excited for this!

PR #642, opened 2022-10-30
for ruby, cache bundle install by default if Gemfile* has not changed

flybayer:
I'm not aware of any cases where `bundle install` uses any files except for Gemfile and Gemfile.lock


coffee-cup:
The one case I know that the `Gemfile` uses other local files is when referencing local gems (The `ruby-local-deps` example does this). We should be able to check this though and include everything if no local deps are included.

For reference this was changed in this PR https://github.com/railwayapp/nixpacks/pull/477 because of an issue with local deps.

src/providers/ruby.rs
```diff
@@ -108,13 +113,12 @@ impl RubyProvider {
...
-            Ok(Some(Phase::build(Some(
-                "bundle exec rake assets:precompile".to_string(),
-            ))))
-        } else {
-            Ok(None)
+            build.add_cmd("bundle exec rake assets:precompile".to_string());
         }
+
+        Ok(Some(build))
     }
```
flybayer, line 122: With the only include Gemfile addition, `COPY . /app/` was never being run for none Rails apps causing an obvious failure.

This change means Ruby apps now always have a build phase, even if no commands are run. This ensures source files get copied.

flybayer:
hey, gentle nudge on this @coffee-cup :)

src/providers/ruby.rs
```diff
@@ -100,6 +100,11 @@ impl RubyProvider {
         let mut install = Phase::install(None);
         install.add_cache_directory(BUNDLE_CACHE_DIR.to_string());
 
+        if !self.uses_gem_dep(app, "local") {
```
coffee-cup, line 103: 👌

coffee-cup (approved):
Thanks! 😄

PR #657, opened 2022-11-04
[Blocked-Ext] Add Bollard Client for Remote Docker Builds

JakeCooper:
This PR: 
- Creates a Docker client struct using Bollard for implementation
- Replaces the Command execs with calls to the Docker client
- Allows attaching a remote docker client

However, due to Bollard not yet supporting BuildKit (See: https://github.com/fussybeaver/bollard/pull/264), we'll wait on that for now.

I've split out the Tokio Async PR so we can merge that (#665)

Currently blocked by https://github.com/fussybeaver/bollard/pull/264


nebulatgs:
I thought we were trying not to be Docker-specific?

JakeCooper:
In an ideal world yes, but we're already coupled. We can remove that coupling at a later time if we want to write interfaces for OCI compliant emiters like Buildah and Podman

I'll switch the clients name from Docker to OCIBuilder or something

JakeCooper:
Okay so, good new and bad news:

Good news it's "done"
Bad news is Bollard doesn't support BuildKit so it doesn't work because of our caching (https://github.com/fussybeaver/bollard/pull/264)

Milo123459:
Not stale

PR #670, opened 2022-11-11
Log app contents when no plan can be generated

coffee-cup:
This PR logs the app directory contents if no build plan can be generated. This should hopefully help make it clearer which files Nixpacks has access to and why no providers were successfully detected.

image


AhmedMozaly (approved):
Awesome! 🎉 
This should make it more clear if the correct directory is being built or not

Issue #621, opened 2022-10-19
Yarn Berry implementation wrongly resolves versions


RDIL:
Hi!

I'm a Railway user, and actually a Yarn team member as of recently.  All my projects use the bleeding edge of Yarn Berry, and the way that the Node Nixpack works is that it is hard-coded to download the latest stable Berry version, as can be seen here:

https://github.com/railwayapp/nixpacks/blob/fac9c30a6f0aa3368a1b648d0398546681dbbb77/src/providers/node/mod.rs#L246

This is **not** what should be done.  As Yarn releases new versions, things can change.
For instance, if a project uses TypeScript, it would have broken with the recent Yarn 3.2.4 release (even though it was a semver-patch release) because the package SHAs changed.

What **should** be done:

- If the `package.json` has a `packageManager` field [(for Corepack)](https://github.com/nodejs/corepack#-corepack), that field should take priority on determining which package manager to use (Corepack should be used to download and run the version).
- If it doesn't, any version specified in the `.yarnrc.yml` should actually *be used*!

Corepack, while still marked as experimental, is the preferred installation method by [Yarn](https://yarnpkg.com/getting-started/install#install-corepack), [pnpm](https://pnpm.io/installation#using-corepack), and also has support for npm.

PR #678, opened 2022-11-15
[Node] Support Corepack properly

RDIL:
Follow-up to #621, #622

#622's implementation works for the most part, but ultimately this is the proper way it should be done


src/providers/node/mod.rs
```diff
@@ -152,6 +159,25 @@ impl NodeProvider {
...
+
+        if package_json.package_manager.is_some() {
+            if package_json
+                .package_manager
+                .expect("Already non-null validated")
+                .starts_with("npm")
+            {
+                // Fall back to just using the system npm version.
+                return Ok(false);
+            }
```
coffee-cup, line 173: Nit: Can use

```rs
if let Some(package_manager) = package_json.package_manger {
  if package_manager.starts_with("npm") {
    // ...
  }
}
```

to avoid using `expect`

src/providers/node/mod.rs
```diff
@@ -95,7 +92,17 @@ impl Provider for NodeProvider {
...
+        let corepack = NodeProvider::uses_corepack(app)?;
+        let mut install = Phase::install(if corepack {
+            Some("npm install -g corepack && corepack enable".to_string())
+        } else {
+            NodeProvider::get_install_command(app)
+        });
+
+        if corepack {
+            install.add_cmd(NodeProvider::get_install_command(app).unwrap());
+        }
```
coffee-cup, line 104: Last nit: Please remove the unwrap here and check if there is actually an install command

RDIL: Oh, didn't see this, will fix.

coffee-cup (approved):
🎉

PR #683, opened 2022-11-19
Fix prisma openssl error

### LINKED ISSUE #574: Support OpenSSL 3
In the latest version of the Nix unstable channel, openssl version 3 is the default instead of version. This breaks a few of our examples that depend on openssl 1 (e.g. node prisma). For now, the nixpkgs archive can be pinned to an older version, but we should do a few things to improve how Nixpacks handles openssl.

- Improved clarity around where openssl comes from. Our base debian image installs openssl with apt-get. However, this version can conflict with openssl from nixpkgs
- Allow users to specify which version of openssl they want. If version 1 is specified, this version should be used for all Nix packages that get installed

https://github.com/NixOS/nixpkgs/compare/a0b7e70db7a55088d3de0cc370a59f9fbcc906c3...8a5b9ee7b7a2b38267c9481f5c629c015108ab0d

coffee-cup:
This PR attempts to upgrade the Nix package archive version, which requires fixing an issue with Prisma clients. The issue occurs because the default openssl version used by all packages (including node) is now openssl 3. But there is some conflict/something that is causing prisma to error in some way.

The current draft of this PR just overrides the nodejs openssl version to 1.1. However, I am not super happy with this and would like Nixpacks to work with openssl 3 across the board. Leaving this PR up as I experiment with different things until we can find a reasonable solution.

## Example errors

image

image

## Things to Try

- [ ] Override the openssl version for the nodejs package
- [ ] Update the openssl version in the base debian image

Fixes #574


Milo123459:
I wonder why this has made the clojure test fail

coffee-cup:
> I wonder why this has made the clojure test fail

I'm guessing that the version of clojure updated? Unsure. I'll take a look at it soon

Milo123459:
Can we see if this fixes the Rust PCR issue?

Issue #630, opened 2022-10-23
Can't use `onlyIncludeFiles` for directories to optimize monorepo builds


flybayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I expected to be able to do something like this to optimize builds by (1) caching npm install and (2) not triggering rebuild by unchanged packages.

```toml
[phases.install]
onlyIncludeFiles = [
  ".npmrc",
  "package.json", 
  "package-lock.json",
  "./packages/shared/package.json", 
  "./packages/shared/package-lock.json",
  "./packages/tower/package.json", 
  "./packages/tower/package-lock.json",
  "./packages/shared/db/schema.prisma"
]
```

But this is using [docker COPY](https://docs.docker.com/engine/reference/builder/#copy) which doesn't copy directories.

### Environment

nixpacks 0.10.4

jhg:
Also this make impossible to add templates dir of rocket as adding `./templates` will copy contents to `/app` instead of to `/app/templates`.

jhg:
I'm doing changes for use `{ from = './templates/index.html.hbs', to = './templates/index.html.hbs' }` syntax while current behavior continues working, an example:

```toml
[start]
  onlyIncludeFiles = [
    '...',
    { from = './templates/index.html.hbs', to = './templates/index.html.hbs' }
  ]
```

Currently changes are in main branch of my fork. I'll open PR soon after add some tests.

PR #685, opened 2022-11-20
Support copy files to a subdirectory in app directory

jhg:
This PR gives a solution for issue #630. Now it's possible to use syntax like:

```toml
[start]
  onlyIncludeFiles = [
    '...',
    { from = './templates/index.html.hbs', to = './templates/index.html.hbs' }
  ]
```

so `index.html.hbs` is copied inside a `templates` subdirectory inside the app directory. That's an example that currently fails when using [Rocket](https://rocket.rs/).

I have an [example of a Rocket service](https://github.com/jhg/simple-parking) that doesn't work on Railway.app because templates' files are copied just in the app directory instead of the template folder.

- [X] Tests are added/updated
- [X] Docs are updated


coffee-cup (commented):
Thanks for this PR! I have briefly looked it over but will have more time to review tomorrow (EST).

However, what are your thoughts on making the default behaviour copy from/to instead of just copying to the app root? It would be a breaking change, but I think that makes the most intuitive sense

flybayer:
I'm definitely in favor of just making `onlyIncludeFiles` work with directories like a normal person would expect :)

jhernandez-at-wiris:
@flybayer Files are copied by Docker, using a `--from=` then I can not know if `./something` is a file without extension or a directory to append to destination or not. I'm not sure if doing `COPY --from=0 ./something {app}/something` will work for files and for directories. If that could work, I can add that change in few minutes. 😄

jhernandez-at-wiris:
I see, I was trying to keep compatibility with current behaviour thinking that maybe it's a problem a breaking change. Then I'll do some breaking changes and that will simplify the changes also. Thx for feedback.

coffee-cup:
In the past we have just never included files from anywhere but the root of the app. I do agree with @flybayer in that the current behavior is unintuitive.

jhernandez-at-wiris:
Maybe also almost uses in the past include the full app directory, that case was not affected by behavior of Docker's `COPY`.

I have the changes done and I'm reviewing tests. In a little I'll update my branch.

src/nixpacks/builder/docker/utils.rs
```diff
@@ -24,28 +24,44 @@ pub fn get_cache_mount(
...
+                let mut file_in_app_dir = app_dir.to_string();
+                if !file_in_app_dir.ends_with('/') {
+                    file_in_app_dir.push('/');
+                }
+                file_in_app_dir.push_str(file.strip_prefix("./").unwrap_or(file));
+                let file = if file.starts_with("./") {
+                    &file_in_app_dir
+                } else {
+                    file
+                };
```
coffee-cup, line 61: Nit: This can be cleaned up a bit by using `Path::join`

```suggestion
                let file_in_app_dir = Path::new(app_dir)
                    .join(file.strip_prefix("./").unwrap_or_default())
                    .display()
                    .to_string();

                let file = if file.starts_with("./") {
                    &file_in_app_dir
                } else {
                    file
                };
```

A similar approach can be used in the above `get_copy_commands`

coffee-cup:
The tests are also failing because the rust musl example is looking for the compiled binary at the incorrect location. That should be a fairly straightforward fix, but lmk if you would like some help updating it!

PR #687, opened 2022-11-21
Add Pipenv support to Python provider

AhmedMozaly:
Some python projects uses [Pipenv](https://pipenv.pypa.io/en/latest/) to manage packages used, This PR adds support to detect `Pipfile` and adjust the install command based on


src/providers/python.rs
```diff
@@ -171,6 +176,20 @@ impl PythonProvider {
 
             install_phase.add_cache_directory(PIP_CACHE_DIR.to_string());
 
+            return Ok(Some(install_phase));
+        } else if app.includes_file("Pipfile") {
+            let cmd = if app.includes_file("Pipfile.lock") {
+                "PIPENV_VENV_IN_PROJECT=1 pipenv install --deploy"
```
coffee-cup, line 182: What does `PIPENV_VENV_IN_PROJECT` do?

src/providers/python.rs
```diff
@@ -207,12 +226,14 @@ impl PythonProvider {
 
     fn is_django(app: &App, _env: &Environment) -> Result<bool> {
         let has_manage = app.includes_file("manage.py");
-        let imports_django = vec!["requirements.txt", "pyproject.toml"].iter().any(|f| {
-            app.read_file(f)
-                .unwrap_or_default()
-                .to_lowercase()
-                .contains("django")
-        });
+        let imports_django = vec!["requirements.txt", "pyproject.toml", "Pipfile"]
```
coffee-cup, line 229: I think this should use `uses_dep` instead (or the implementation of `uses_dep` should change to match this)

coffee-cup (approved):
suggestion: add a docker run test too.

But overall it looks good!

src/providers/python.rs
AhmedMozaly: by default Pipenv creates an environment directory in some random location (for example `/root/.local/share/virtualenvs/app-4PlAip0Q`).

`PIPENV_VENV_IN_PROJECT` tells it that there is an already activated `venv` environment, So Pipenv will use the same directory instead of creating new one (in our case it's ` /app/.venv`)

PR #709, opened 2022-12-13
[WIP] Remote Docker support via Bollard

### LINKED ISSUE #564: Rust-based Docker API client instead of relying on `docker` command
### Feature request

Using a Rust-based Docker API client would remove the dependency on a local Docker client (provided by the `docker` CLI). This makes it easier to use a remote Docker daemon without ever having installed `docker` locally.

### Motivation

At Fly.io, we use "remote" docker builders so users without Docker running can still build images for our platform.

Looks like `nixpacks` works for us even if the Docker daemon isn't running, but users still need the `docker` CLI binary available in their `PATH`.

### Contribution

I could submit a PR, but I'm not sure there's a good non-async Docker client and without a bit more guidance on how to proceed I wouldn't want to sink time in a big "add `async` stuff to `nixpacks`" effort :smile:

JakeCooper:
This PR adds Bollard for async docker support

Fixes #564 

Having an issue with "no active sessions". Filed https://github.com/fussybeaver/bollard/issues/280


jeromegn:
Uh oh, this seemed promising!

PR #718, opened 2022-12-15
Add traits Send And Sync to the Provider type

WindowsCmd:
This adds support for running `create_docker_image` inside of tokio threads.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,7 @@ mod chain;
 pub mod nixpacks;
 pub mod providers;
 
-pub fn get_providers() -> &'static [&'static dyn Provider] {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -47,7 +47,7 @@ mod chain;
 pub mod nixpacks;
 pub mod providers;
 
-pub fn get_providers() -> &'static [&'static dyn Provider] {
+pub fn get_providers() -> &'static [&'static (dyn Provider + Send + Sync)] {
```
coffee-cup, line 50: Can these traits be added onto the main `Provider` struct instead of here?

Milo123459:
please make a merge commit to bring in the clippy fixes

coffee-cup (approved):

PR #727, opened 2022-12-17
Install ruby with rbenv

coffee-cup:
This PR updates the Ruby provider to install Ruby with [rbenv](https://github.com/rbenv/rbenv) instead of rvm.

Rbenv has a few benefits over rvm
- Handles installing Openssl 1.1 automatically for Ruby versions <3.1.0
- Installs in about half the time as it does for rvm 134s vs 269s (shocked how close that is to exactly 50% tbh)


JakeCooper (approved):

JakeCooper (approved):
LGTM should merge when you're good. Ruby support seems abysmal right now (trying to get Discourse up for a client)

jarthod:
For the record you can reduce the build time even more by disabling the ruby documentation generation:
```toml
# nixpacks.toml
[variables]
RUBY_CONFIGURE_OPTS="--disable-install-rdoc"
```
In my test I went from 150sec to 117sec for ruby 2.6.9.
Maybe this should be set by default? I don't see any reason to generate the documentation for a production container.

---

In my case I also use [jemalloc to reduce memory](https://medium.com/@adrienjarthon/ruby-jemalloc-results-for-updown-io-d3fb2d32f67f) usage with:
```toml
# nixpacks.toml
[variables]
RUBY_CONFIGURE_OPTS="--with-jemalloc --disable-install-rdoc"

[phases.setup]
aptPkgs = ["...", "libjemalloc-dev"]
```
Though I don't think this should be by default as it would probably break with other ruby implementations (non MRI).

Issue #701, opened 2022-12-03
PNPM Workspaces with NX do not work


nebulatgs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The generated start command is `pnpm --workspace ${WORKSPACE} run start`

### Expected behavior

This is invalid, and it should be `pnpm --filter ${WORKSPACE} run start`

### Environment

Railway - Nixpacks 0.15.0

PR #733, opened 2022-12-20
PNPM fixes

### LINKED ISSUE #702: PNPM Cache Location is incorrect

### Describe the bug

Nixpacks attempts to cache `/root/.cache/pnpm`, but this is incorrect

### To reproduce

_No response_

### Expected behavior

The correct store directory is `/root/.local/share/pnpm/store/v3`

### Environment

Railway - Nixpacks 0.15.0

Milo123459:
This PR fixes the PNPM cache directory and the way NX apps are started with PNPM. This also contains some general nx fixes which swaps from using `npx` always to the package manager dlx command.

Fixes #702, #701

### DIFF
--- a/src/providers/node/nx.rs
+++ b/src/providers/node/nx.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 use crate::nixpacks::{app::App, environment::Environment};
+use crate::providers::node::NodeProvider;
 
 #[derive(Debug, Serialize, PartialEq, Eq, Deserialize)]
 pub struct NxJson {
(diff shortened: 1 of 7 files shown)


src/providers/node/nx.rs
```diff
@@ -106,10 +111,16 @@ impl Nx {
             if let Some(start_target) = project_json.targets.start {
                 if let Some(configurations) = start_target.configurations {
                     if configurations.production.is_some() {
-                        return Ok(Some(format!("npx nx run {nx_app_name}:start:production")));
+                        return Ok(Some(format!(
+                            "{} nx run {nx_app_name}:start:production",
+                            NodeProvider::get_package_manager_dlx_command(app)
```
coffee-cup, line 116: This is fine for now, but we likely want to refactor the Node provider soon since this creates a circular dependency. Node -> Nx -> Node

coffee-cup (approved):

PR #755, opened 2023-01-10
allow insecure nix package installs

coffee-cup:
This PR allows insecure Nix package installs. This is needed to install Python 2.7, which has reached its end of life.

image


Milo123459 (approved):
I guess this fixes #753?

PR #757, opened 2023-01-10
update ubuntu base image on schedule

coffee-cup:
This PR updates the Nixpacks ubuntu base image as part of the `publish_ubuntu` workflow


Milo123459 (approved):
Is this the same way it's done on the Debian image?

coffee-cup:
> Is this the same way it's done on the Debian image?

Yup!

PR #762, opened 2023-01-10
Ruby Version Envvar

JakeCooper:
This PR consumes the Ruby environment variable version from an environment variable if it exists

### DIFF
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -28,7 +28,7 @@ impl Provider for RubyProvider {
 
     fn get_build_plan(&self, app: &App, env: &Environment) -> Result<Option<BuildPlan>> {
         let setup = self.get_setup(app, env)?;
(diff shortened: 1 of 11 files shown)


src/providers/ruby.rs
```diff
@@ -323,7 +337,25 @@ mod test {
...
+
+    #[test]
+    fn test_version_arg() -> Result<()> {
+        assert_eq!(
+            RubyProvider::get_ruby_version(
+                &RubyProvider {},
+                &App::new("./examples/ruby")?,
+                &Environment::new(BTreeMap::from([(
+                    "NIXPACKS_RUBY_VERSION".to_string(),
+                    "3.1.2".to_string()
```
coffee-cup, line 357: nit: Change this version since `3.1.2` is used when the env var is not provided

coffee-cup (approved):

Issue #372, opened 2022-07-29
Support Clojure deps.edn


Sohalt:
### Feature request

It would be nice to support Clojure projects based on [deps.edn](https://clojure.org/guides/deps_and_cli) and [tools.build](https://clojure.org/guides/tools_build).
It's not exactly easy, since there are basically no constraints to what you can do in a `deps.edn` and how your build process looks, but it would be nice to have some heuristics to cover the most common use-cases (e.g. sth like `clojure -X:build`).

### Motivation

Currently only Clojure projects based on [leiningen](https://leiningen.org/) are supported.

### Contribution

I might look into this myself, if I find the time. I'll leave a comment here, before I start any work. In the meantime, if someone else wants to have a go, I'd be very grateful.

AhmedMozaly:
Yes, deps.edn support is planned
Would also leave a comment when started!

AhmedMozaly:
BTW, do you have reference repo to use while testing deps.edn support?

PR #771, opened 2023-01-24
Add support for Clojure `tools.build` projects

rads:
The [`tools.build`](https://clojure.org/guides/tools_build) library is the official library for building Clojure projects from a `deps.edn` file. This PR uses the `build.clj` file for detection since technically other libraries besides `tools.build` could be built on top of `deps.edn` in the future.

Resolves https://github.com/railwayapp/nixpacks/issues/372.

### DIFF
--- a/src/providers/clojure.rs
+++ b/src/providers/clojure.rs
@@ -21,22 +21,23 @@ impl Provider for ClojureProvider {
     }
 
     fn detect(&self, app: &App, _env: &Environment) -> Result<bool> {
-        Ok(app.includes_file("project.clj"))
+        Ok(self.is_using_lein(app) || self.is_using_tools_build(app))
     }
 
     fn get_build_plan(&self, app: &App, env: &Environment) -> Result<Option<BuildPlan>> {
         let setup = Phase::setup(Some(vec![
-            Pkg::new("leiningen"),
+            if self.is_using_tools_build(app) {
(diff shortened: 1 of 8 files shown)


ndneighbor:
Will defer to @coffee-cup on what Nixpacks version this will go in. Also, it appears that CI needs to be appeased.

rads:
@ndneighbor: Thanks for the quick review! I think I fixed the CI failures now.

ndneighbor (approved):

src/providers/clojure.rs
```diff
@@ -54,6 +55,23 @@ impl Provider for ClojureProvider {
 }
 
 impl ClojureProvider {
+    fn has_lein_ring_plugin(&self, app: &App) -> bool {
+        self.is_using_lein(app)
+            && app
+                .read_file("project.clj")
+                .unwrap()
```
coffee-cup, line 62: Don't explicitly unwrap incase the `project.clj` file does not exist
```suggestion
                .unwrap_or_default()
```

coffee-cup (approved):
Looks good! Just a small comment about unwrapping

src/providers/clojure.rs
rads: Thanks, I'm a total Rust newbie. 😄

Milo123459:
Is this good to merge?

PR #785, opened 2023-02-09
Improve NX docs

flybayer:
this improves the NX docs and adds missing info on using `NIXPACKS_NX_APP_NAME` env var


coffee-cup (approved):
🎉

PR #787, opened 2023-02-10
Rust: support top-level main bin

coffee-cup:
If there are multiple bins in a `src/bin` directory, we did not also support the top-level `main.rs` file as a bin. This PR fixes that

### DIFF
--- a/examples/rust-multiple-bins/src/main.rs
+++ b/examples/rust-multiple-bins/src/main.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("This is the main Bin!");
+}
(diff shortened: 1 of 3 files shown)


JakeCooper (approved):

Milo123459:
this supports default_run right

coffee-cup:
> this supports default-run right

Yup. It supported it before too. The problem was just that the main.rs bin was not be included

Milo123459 (approved):

Milo123459:
is there anything blocking this from being merged?

PR #789, opened 2023-02-11
feat(scala): reduced scala sbt image size

### LINKED ISSUE #788: Reduce sbt scala jvm image size
### Feature request

Hi looks like the scala bundle sizes are quite large. I feel we could certainly trim it by using multi stage builds, where the run phase only has `jre_minimal` and not sbt. I was wondering if we could install specific packages in the start/run phase like we did in the setup phase.

In the future we can also add graal support. I would like to trim jvm image for now

### Motivation

Reduce image size

### Contribution

I will submit a PR for this

yzia2000:
The Start phase runs the scala jar in a lean eclipse temurin 17.0.5 (official adoptium) image with only the package files. This reduced the image size from 2GB to ~200mb for the example project.

Resolves #788

### DIFF
--- a/src/providers/scala.rs
+++ b/src/providers/scala.rs
@@ -44,9 +44,14 @@ impl Provider for ScalaProvider {
             build.add_cache_directory("/root/.cache/coursier");
             build.depends_on_phase("setup");
 
-            let start_cmd = self.get_start_cmd(app).map(StartPhase::new);
+            let start_phase = self.get_start_cmd(app).map(StartPhase::new).map(|phase| {
+                let mut updated_phase = phase;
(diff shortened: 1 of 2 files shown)


src/providers/scala.rs
```diff
@@ -76,6 +81,11 @@ impl ScalaProvider {
         pkgs
     }
 
+    pub fn get_run_dep_pkgs(&self) -> Vec<Pkg> {
```
Milo123459, line 84: this function isn't used anywhere?

yzia2000: ah I forgot to remove that. Was trying to see if I can use a nix image in the StartPhase and packages to it. let me remove it.

coffee-cup (approved):

PR #795, opened 2023-02-16
Fix SDK version parsing

Milo123459:
This PR fixes SDK version parsing. Previously, it would spit out incorrect package names.

### DIFF
--- a/src/providers/csharp.rs
+++ b/src/providers/csharp.rs
@@ -120,7 +120,7 @@ mod test {
 
     #[test]
     fn test_global_json() -> Result<()> {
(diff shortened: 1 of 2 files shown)


coffee-cup (approved):

MikePreston17 (commented):
If your language supports Regex and Named Capture groups, you won't have to hardcode or use a fluid DSL here.  The regex wouldn't be that complicated and Microsoft has a table of all the current sdks (2, at the time of this writing).

Maybe this will give you an idea.  That's how I'd do it, anyways.  No more updating the magic strings - only updating the pattern, if it ever changes.  [Xregexp](https://xregexp.com/) is a good example of what I mean.

PR #797, opened 2023-02-16
Support nested directories

Milo123459:
This PR allows Nixpacks to search sub directories if no plan can be generated. This only happens if one sub directory is present. This will cut down a significant amount of help posts we get in the Railway discord.


src/lib.rs
```diff
@@ -105,11 +105,26 @@ pub async fn create_docker_image(
...
+    let sub = &app.paths.get(0).unwrap().display().to_string()[len..];
+    if sub.split(std::path::MAIN_SEPARATOR).count() > 1 {
+        let mut path = sub.split(std::path::MAIN_SEPARATOR).collect::<Vec<&str>>();
+        path.remove(path.len() - 1);
+        path.remove(0);
+        println!(
+            "Using subdirectory \"{}\"",
+            path.join(std::path::MAIN_SEPARATOR.to_string().as_str())
+        );
+    }
```
coffee-cup, line 127: This can be cleaned up a bit by using `Path::strip_prefix`

```rs
  let app = App::new(path)?;
    let environment = Environment::from_envs(envs)?;

    let orig_path = app.source.clone();

    let mut generator = NixpacksBuildPlanGenerator::new(get_providers(), plan_options.clone());
    let (plan, app) = generator.generate_plan(&app, &environment)?;

    if let Ok(subdir) = app.source.strip_prefix(orig_path) {
        if subdir != std::path::Path::new("") {
            println!(
                "Using subdirectory \"{}\"",
                subdir.to_str().unwrap().to_string()
            );
        }
    }
```

Milo123459: this doesnt seem to work correctly

coffee-cup (approved):

PR #798, opened 2023-02-16
Fix cobol provider

FinnDore:
This pr fixes the cobol provider that broke as a result of the nix archive update.

Not to sure if ive done git propper though. guess we will see once i make this pr...


FinnDore:
Leme fix the archive discrepancy

Milo123459 (approved):

FinnDore:
fuck it we go again

PR #807, opened 2023-02-23
Update default node version to 18 

wyzlle:
This PR updates default node version from 16, which is in maintenance, to version 18, which has active long-term support (LTS).
image


coffee-cup (changes requested):
Marking this as _request changes_ so it doesn't accidentally get merged.

See this discussion on Discord for more info https://canary.discord.com/channels/713503345364697088/1078123635526610944/1078135352612626533

PR #816, opened 2023-03-01
Ruby: Environment variables and PATH addition

henrikbjorn:
that is the default for bundle binstubs

This is based on the environment variables added by Heroku in their buildpacks.

Also it adds `/app/bin` to the path like the `NodeProvider` does for `node_modules/.bin`.

Another change I would like to make, is to split up the `rbenv` commands into one line per command instead of the long and unreadable `&&` cmd

### DIFF
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -119,6 +119,7 @@ impl RubyProvider {
         ));
 
         setup.add_path("$HOME/.rbenv/bin".to_string());
+        setup.add_path("/app/bin".to_string());
 
         Ok(Some(setup))
     }
@@ -169,6 +170,9 @@ impl RubyProvider {
         let ruby_version = self.get_ruby_version(app, env)?;
         let mut env_vars = EnvironmentVariables::from([
             ("BUNDLE_GEMFILE".to_string(), "/app/Gemfile".to_string()),
+            ("BUNDLE_WITHOUT".to_string(), "test:development".to_string()),
+            ("BUNDLE_CLEAN".to_string(), "1".to_string()),
+            ("BUNDLE_DEPLOYMENT".to_string(), "1".to_string()),
             (
                 "GEM_PATH".to_string(),
                 format!(
@@ -183,6 +187,8 @@ impl RubyProvider {
         ]);
 
         if self.is_rails_app(app) {
+            env_vars.insert("RAILS_ENV".to_string(), "production".to_string());
+            env_vars.insert("SECRET_KEY_BASE".to_string(), "1".to_string());
             env_vars.insert("RAILS_LOG_TO_STDOUT".to_string(), "enabled".to_string());
             env_vars.insert("RAILS_SERVE_STATIC_FILES".to_string(), "1".to_string());
         }
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -130,8 +130,12 @@ impl RubyProvider {
 
         if !self.uses_gem_dep(app, "local") {
             // Only run install if Gemfile or Gemfile.lock has changed
-            install.only_include_files =
-                Some(vec!["Gemfile".to_string(), "Gemfile.lock".to_string(), "vendor/cache".to_string()]);
+            install.add_file_dependency("Gemfile".to_string());
+            install.add_file_dependency("Gemfile.lock".to_string());
+
+            if app.includes_file("vendor/cache") {
+              install.add_file_dependency("vendor/cache".to_string())
+            }
         }
 
         install.add_cmd("bundle install".to_string());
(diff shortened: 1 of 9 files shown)


src/providers/ruby.rs
```diff
@@ -183,6 +187,8 @@ impl RubyProvider {
         ]);
 
         if self.is_rails_app(app) {
+            env_vars.insert("RAILS_ENV".to_string(), "production".to_string());
+            env_vars.insert("SECRET_KEY_BASE".to_string(), "1".to_string());
```
henrikbjorn, line 191: This is required for asset compilation in production. I assume this will be overriden with the env defined by Railway in their UI

henrikbjorn:
Other changes could be to copy `vendor/cache` since it will speed up bundle install as it will using the local cached gems in that directory

src/providers/ruby.rs
coffee-cup: Forgive me as I am not that familiar with deploying Ruby apps. But is it common for this env var to have a default value? Seems like it could be a security risk with someone forgetting to override this value when deploying to production.

coffee-cup (commented):
Thanks for the PR! This is looking good. Just one comment about the `SECRET_KEY_BASE` variable.

Could you also please update the snapshot tests https://github.com/railwayapp/nixpacks/blob/main/CONTRIBUTING.md#snapshot-tests

src/providers/ruby.rs
henrikbjorn: When precompiling assets a `SECRET_KEY_BASE` needs to be present. It might not be a good thing to not remember changing.

However we could inline it in the `bundle exec rails assets:precompile` call ?

henrikbjorn:
```
~/P/nixpacks ❯❯❯ cargo insta test --review -- --test generate_plan_tests

   Compiling nixpacks v1.4.0 (/Users/Henrik/Projects/nixpacks)
    Finished test [unoptimized + debuginfo] target(s) in 3.73s
     Running unittests src/lib.rs (target/debug/deps/nixpacks-08f0fe7512a69dc8)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 85 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/debug/deps/nixpacks-3c88388bea1c86aa)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/docker_run_tests.rs (target/debug/deps/docker_run_tests-857c689699b69091)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 69 filtered out; finished in 0.00s

     Running tests/generate_plan_tests.rs (target/debug/deps/generate_plan_tests-8a0d05552663b61d)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 103 filtered out; finished in 0.00s

   Doc-tests nixpacks

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

done: no snapshots to review
```

Nothing happens when I run the snapshot commands from the `CONTRIBUTING.md` file.

henrikbjorn:
Changed the adding of `vendor/cache` and the gem files to use `add_file_dependency` which seems to be method used by the other providers.

src/providers/ruby.rs
henrikbjorn: Maybe it should be possible to add build time environment variables, but that might increase complexity of nixpacks

henrikbjorn: Seems that Heroku just inject it also https://github.com/heroku/buildpacks-ruby/blob/d988c4371cc5aef11e1c7375fad4f7ed87475de8/buildpacks/ruby/src/layers/bundle_install_layer.rs#L435

coffee-cup: What you linked is just a test. For prod it looks like they generate a random one https://github.com/heroku/buildpacks-ruby/blob/d988c4371cc5aef11e1c7375fad4f7ed87475de8/buildpacks/ruby/src/steps/default_env.rs#L51

I think it would be a good idea for Nixpacks to do the same. Thoughts?

coffee-cup:
Hmm that is odd. You can also use the shorthand `cargo snapshot`

I see this when I run it

image

src/providers/ruby.rs
henrikbjorn: Ohh yeah

It seems it inserts a random one for the first build, but then uses an existing one otherwise. Is this possible for nixpacks? To generate an ENV and re-use it for the next build?

henrikbjorn:
@coffee-cup I can't run the tests locally also. I think it have something to do with me being on a M1 Arm processor

src/providers/ruby.rs
coffee-cup: At the moment, it is not possible. Nixpacks builds each image isolated. We would have to return some information that is saved in the host platform. In the meantime, I think we should remove the automatic `SECRET_KEY_BASE` variable.

PR #817, opened 2023-03-02
Better Ruby caching

Milo123459:
This PR caches the .rbenv directory. This will lead to significant speedups in Ruby builds.

### DIFF
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -112,8 +112,14 @@ impl RubyProvider {
 
         setup.add_cmd(format!(
             "curl -fsSL https://github.com/rbenv/rbenv-installer/raw/HEAD/bin/rbenv-installer | bash -s stable \
-            && printf '\\neval \"$(rbenv init -)\"' >> /root/.profile \
-            && . /root/.profile \
+            && printf '\\neval \"$(rbenv init -)\"' >> $HOME/.profile \
+            && sed -i 's_PATH=$HOME/.rbenv/bin:$PATH_PATH=$HOME/keep/rbenv/bin:$PATH_' $HOME/.profile \
+            && sed -i 's_rbenv init_$HOME/keep/rbenv/bin/rbenv init_' $HOME/.profile \
+            && mkdir $HOME/keep \
+            && mkdir $HOME/keep/rbenv \
+            && cd $HOME/.rbenv \
+            && cp . $HOME/keep/rbenv -R \
+            && . $HOME/.profile \
             && rbenv install {ruby_version} \
             && rbenv global {ruby_version} \
             && gem install {bundler_version}"
@@ -134,7 +140,6 @@ impl RubyProvider {
             install.only_include_files =
                 Some(vec!["Gemfile".to_string(), "Gemfile.lock".to_string()]);
         }
-
         install.add_cmd("bundle install".to_string());
 
         // Ensure that the ruby executable is in the PATH
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -142,7 +142,8 @@ impl RubyProvider {
         install.add_path(format!("/usr/local/rvm/rubies/{ruby_version}/bin"));
         install.add_path(format!("/usr/local/rvm/gems/{ruby_version}/bin"));
         install.add_path(format!("/usr/local/rvm/gems/{ruby_version}@global/bin"));
-
+        install.add_cmd("bundle clean --force".to_string());
+        
         Ok(Some(install))
     }
 
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -14,6 +14,7 @@ use regex::Regex;
 pub struct RubyProvider {}
 
 const BUNDLE_CACHE_DIR: &str = "/root/.bundle/cache";
+const RBENV_CACHE_DIR: &str = "/.rbenv";
 
 impl Provider for RubyProvider {
     fn name(&self) -> &str {
@@ -126,6 +127,7 @@ impl RubyProvider {
     fn get_install(&self, app: &App, env: &Environment) -> Result<Option<Phase>> {
         let mut install = Phase::install(None);
         install.add_cache_directory(BUNDLE_CACHE_DIR.to_string());
+        install.add_cache_directory(RBENV_CACHE_DIR.to_string());
 
         if !self.uses_gem_dep(app, "local") {
             // Only run install if Gemfile or Gemfile.lock has changed
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -143,7 +143,7 @@ impl RubyProvider {
         install.add_path(format!("/usr/local/rvm/gems/{ruby_version}/bin"));
         install.add_path(format!("/usr/local/rvm/gems/{ruby_version}@global/bin"));
         install.add_cmd("bundle clean --force".to_string());
-        
+
         Ok(Some(install))
     }
 
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -14,7 +14,7 @@ use regex::Regex;
 pub struct RubyProvider {}
 
 const BUNDLE_CACHE_DIR: &str = "/root/.bundle/cache";
-const RBENV_CACHE_DIR: &str = "/.rbenv";
+const RBENV_CACHE_DIR: &str = "/root/.rbenv";
 
 impl Provider for RubyProvider {
     fn name(&self) -> &str {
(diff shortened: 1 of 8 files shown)


jakubzelazny:
Thanks for the PR, Milo 🙌

Little explanation: when using rbenv, the gems are located in `/.rbenv/versions/{ruby-version-number}/lib/ruby/gems/{ruby-minor-version}`. Hence the suggested cache path.

This article goes into detail: https://www.honeybadger.io/blog/rbenv-rubygems-bundler-path/

jakubzelazny:
Not sure if it would help. The path to specific gems is:
`/.rbenv/versions/{ruby-version-number}/lib/ruby/gems/{ruby-minor-version}/<gem-version-and-name>`

For example: `/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/pay-6.3.1`

What kind of problems are you thinking about @henrikbjorn?

henrikbjorn:
> 

What happens if I switch from `3.1.0` to `3.2.0` then the cache would be reused but the contents would be invalid.

Generally I just think being more precise is the better option e.g. `.rbenv/versions/3.1.0`

jakubzelazny:
Agreed. Not sure how you cache works but what happens when the version of one of  the gems changes? Will you invalidate cache somehow?

For example from
`.../3.1.0/gems/pay-6.3.1`
to
`.../3.1.0/gems/pay-7.0.1`

henrikbjorn:
> 

I don't really know either (very new here and to nixpacks). I guess my worries are not really an issue

henrikbjorn:
> > 
> 
> I don't really know either (very new here and to nixpacks). I guess my worries are not really an issue

Usually we use `vendor/cache` for the gems and then it is very easy for the install of gems. And I would recommend that over fetching them when building.

jakubzelazny:
Bundler has a command which creates cache in `vendor/cache`.
It's `bundle package`.

From the docs
```Copy all of the .gem files needed to run the application into the vendor/cache directory. In the future, when running bundle-install, use the gems in the cache in preference to the ones on rubygems.org.```

https://bundler.io/v2.0/man/bundle-package.1.html

**Comment by henrikbjorn** [comment]
That is what I meant. We use that for our rails applications :)

**Comment by jarthod** [comment]
Jumping in here as I just noticed this same issue.

@henrikbjorn if we use just `.rbenv` in case of ruby upgrade the cache would be reused indeed but would not cause any problem because it'll re-install gems in a different sub folder anyway, it's the same as installing a new ruby version locally while still having gems in other versions. It would waste some space though, because the gems cached in a previous folder would stay in cache uselessly.

If we cache only `.rbenv/versions/3.1.1` then we only cache the current version, but what would happen to the previouly cached `.rbenv/versions/3.1.0` folder?  if it's still in cache and restored every time it won't help much?

Also doing this (`.rbenv` or `.rbenv/versions/3.1.1`) would not only cache gems but the whole ruby installation, which could actually be very nice if we could restore it before re-installing ruby every time (would save a LOT of time and avoid wasting ressources). But maybe there's a good reason ruby is recompiled every time ?

Also by the way rbenv is in `/root/.rbenv` apparently:
```
#11 27.86 Installed openssl-1.1.1t to /root/.rbenv/versions/2.6.9
#11 110.2 Installed ruby-2.6.9 to /root/.rbenv/versions/2.6.9
```

If we **only** want to cache **the gems**, I think we should use ` /root/.rbenv/versions/{ruby-version-number}/lib/ruby/gems`

Though I tried doing this from `nixpacks.toml` with:
```toml
[phases.setup]
cacheDirectories = ['/root/.rbenv/versions/2.6.9/lib/ruby/gems']
```
Or
```toml
[phases.setup]
cacheDirectories = ['/root/.rbenv']
```
And in both case it does not work because the cache folder is unmounted after the build, so all gems are gone and ruby is gone for the best steps and the runtime. As this caching mechanism only works for temporary builds files, it can't help with ruby or gem build time, unless we add another step to copy the resulting files.

> Usually we use vendor/cache for the gems and then it is very easy for the install of gems. And I would recommend that over fetching them when building.

IMO if we only cache the gems this is a better idea yes. Specify a custom bundle folder during `bundle install` and cache this one. If we can cache ruby too to make build faster though, then it may make sense to cache rbenv entirely. But that cannot work
with the existing caching mechanism.

> Bundler has a command which creates cache in vendor/cache. It's `bundle package`.

Indeed but this only saves the network download because all gems will still be re-installed / re-build every time. For gems with compiled extensions this is what takes time. In order to install the gems in a different folder (by default vendor/bundle) it's `--deployment` which is usually used in production (by capistrano for example). So the only way to really save time would be to cache the *installed* gems folder, and ideally the ruby install. But I'm not sure how to do that with nixpacks at the moment.

Milo123459:
The nixpacks cache always updates, and is always retrieved. If you have multiple versions, it doesn't matter, and it will stay up to date.

henrikbjorn:
@Milo123459 But won't it increase over time then? Accumulating old gems as they are upgraded?

henrikbjorn:
@Milo123459 There is a `bundle clean` which will handle that case. That is also a default for Heroku via `BUNDLE_CLEAN=1`. However it breaks on my system as nixpacks runs everything as root and then it want's a confirmation for running it. Since it cleans the system gems.

Milo123459:
So, I could run bundle clean in the install phase.

Milo123459:
Should work now. Thanks for pointing this out, I thought `bundle install` would auto remove not-needed ones. BTW, Just to check, are you sure gems don't get installed into the bundle cache dir at "/root/.bundle/cache"?

henrikbjorn:
I also tried setting some environment variables in my PR and add `vendor/cache` as a cache directory for `bundle package` packages apps.

On my M1 I can't run the tests etc. So maybe you can include it here. The changes are small

jarthod:
But how does this actually help as the `.rbenv` folder is not being used (it's `/root/.rbenv` in my testing) ?

Milo123459:
Chaning cache path to /root/.rbenv (which is where the directory is) results in the rbenv command not being found..

henrikbjorn:
I think there should be a refactor or rethink to the ruby provider. How are rbenv installing gems, what is BUNDLE_CACHE doing etc.

Not sure what the rvm code in the ruby provider does.

jarthod:
> Chaning cache path to /root/.rbenv (which is where the directory is) results in the rbenv command not being found..

Yes, because now it's working and thus removing the folder at the end. feel free to read my previous comment.
You can't fix this problem that easily.

jarthod:
This is by design apparently: https://nixpacks.com/docs/configuration/caching

> The contents of these directories are restored before the install/build phases are run and **cleared afterwards**

That's what I analyzed in my first comment

jarthod:
For the record I spent some time trying to optimize nixpacks default ruby setup (maybe starting from existing work like Heroku buildpacks could have helped here, feel like reinventing the wheel ^^).
Here's the toml config I came up with:
```nixpacks.toml
[variables]
RAILS_ENV="production"
# disable documentation compilation in ruby for faster build and smaller footprint (also mentionned in #727)
# build ruby with jemalloc for lower memory footprint (optional)
RUBY_CONFIGURE_OPTS="--with-jemalloc --disable-install-rdoc"

# Result for my project:
# - 15-20 sec to install apt packages
# - 100 sec to install ruby 2.6.9 (no doc, including 20 sec to install openssl 1.1.1)
# - 26 sec to install gems
# - 5 sec for assets:precompile
# Total: ~150, but 40sec with ruby and APT cached

[phases.setup]
aptPkgs = ["...", "libjemalloc-dev"]
# Below is an attempt to cache ruby and gems (works for ruby only as gems seems to be installed between the two steps so they don't benefit from the caching)
# Had to rewrite all commands here to change `rbenv install -s` and `rbenv global 2.6.9 && gem install bundler:2.3.21 --no-document --conservative` and add the `cp` commands to copy the cached .rbenv folder back and forth to circumvent the caching volatility. I couldn't find a way to interpolate Ruby version and bundler version in here.
cmds = ["cp -r /cache/.rbenv /root", "curl -fsSL https://github.com/rbenv/rbenv-installer/raw/HEAD/bin/rbenv-installer | bash -s stable && printf '\\neval \"$(rbenv init -)\"' >> /root/.profile && . /root/.profile && rbenv install -s 2.6.9 && rbenv global 2.6.9 && gem install bundler:2.3.21 --no-document --conservative", "cp -r /root/.rbenv /cache"]
cacheDirectories = ["/cache/.rbenv"]

[phases.build]
cacheDirectories = ["/cache/.rbenv"]
cmds = ["cp -r /cache/.rbenv /root", "...", "cp -r /root/.rbenv /cache"]
```

This may be helpful if you want to optimize the defaults (or for other users in the meantime).

jakubzelazny:
@jarthod that's a significant improvement. Thanks for the toml config. I'll try it on my project later.

Milo123459:
We can merge this PR when we have more fine control over caching directories, and along with that, we can add more improvements (ie caching gems).

jpowell:
Any update on this? I've got two parallel builds on Railway and Render, and Render smokes Railway bad...

Issue #841, opened 2023-03-24
Create a Cachix (or other) binary cache for the nixpacks flake


piperswe:
### Feature request

Nixpacks can be installed using the flake in this repository:

```bash
nix profile install github:railwayapp/nixpacks/v1.5.1#nixpacks
```

Installing this way requires building the crate, which can take quite a while. It would be great if there were a binary cache I could use to avoid having to build from scratch. This could be in the form of a Cachix cache which is pushed to from a [GitHub action](https://nix.dev/tutorials/continuous-integration-github-actions). Alternatively, you could add [garnix](https://garnix.io/) to this repo, which automatically builds all outputs for a repo's flake and uploads them to their binary cache.

### Motivation

I include Nixpacks in my [devenv](https://devenv.sh), and that devenv is used in my CI pipeline. Building Nixpacks from scratch can add quite a bit of time to my CI. I could build and upload to a binary cache of my own through my personal [Hydra instance](https://hydra.piperswe.me), but it would benefit others to have an official binary cache, and this repo's flake doesn't contain any checks my Hydra instance can build.

### Contribution

I'll put in a PR to add checks to the flake that my (and others') Hydra instance(s) can build. From there, as someone not affiliated with Railway, I don't know as I can really contribute much to this.

PR #843, opened 2023-03-24
Improve Nix flake by adding checks, auto-updating version, and fixing build

piperswe:
The checks will allow the flake to be built using Hydra. It looks like this project is released with `cargo release`, so I added a config file that will automatically update the flake's version. The package build wasn't running for a variety of reasons, so I reconfigured it to hopefully be more stable in the future.

Helps with #841.

### DIFF
--- a/release.toml
+++ b/release.toml
@@ -1,0 +1,4 @@
+# Automatically update version in flake.nix on `cargo release`
+pre-release-replacements = [
+  { file="flake.nix", search="version = \".*\";", replace="version = \"{{version}}\";" },
+]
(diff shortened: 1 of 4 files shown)


release.toml
```diff
@@ -0,0 +1,4 @@
+# Automatically update version in flake.nix on `cargo release`
+pre-release-replacements = [
```
coffee-cup, line 2: TIL this was possible :D

coffee-cup (approved):
Thanks!

release.toml
piperswe: Same here! Found it with some Googling

PR #850, opened 2023-03-30
Store default arguments in generated Dockerfile

joshraker:
This PR updates Dockerfile generation to include default `ARG` values.

I'm currently working on a project that uses `nixpacks build --out` to generate Dockerfiles but I came to realize that when I built the images myself they didn't work. The culprit turned out to be the missing `--build-arg`s that `nixpacks` supplies when it builds the image. I made this change on my local fork and my images started working without my having to supply `--build-arg`s and figured others using the `--out` argument may find value in the Dockerfile working without any additional input.

### DIFF
--- a/src/nixpacks/builder/docker/dockerfile_generation.rs
+++ b/src/nixpacks/builder/docker/dockerfile_generation.rs
@@ -143,7 +143,7 @@ impl DockerfileGenerator for BuildPlan {
                 // Pull the variables in from docker `--build-arg`
                 variables
                     .iter()
-                    .map(|var| var.0.to_string())
+                    .map(|var| format!("{}={}", var.0, var.1))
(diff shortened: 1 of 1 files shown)


coffee-cup (commented):
Hmm I think there was a specific reason it was just `ARG {value}` before as opposed to embedding the variable in the Dockerfile. But I can't remember why 🤔 
I think this looks good though. I will test it thoroughly and can merge or close soon :)

coffee-cup:
> I suppose if there's a situation where secrets are being passed in as --env variables then they would end up in the resulting Dockerfile and, potentially, the build logs. Was it related to that?

Yes it is exactly this. I think splitting the behaviour is a good idea so that we are not accidentally leaking secrets. Nixpacks doesn't do anything special when running the Dockerfile to supress them

PR #855, opened 2023-04-06
fix "Failed to find django application name" error

flybayer:
Previously Django applications without wsgi would totally fail with `"Failed to find django application name"` and no way to get around it.

This fixes it, by falling back to `python manage.py migrate && python manage.py runserver`


flybayer:
alternate change: https://github.com/railwayapp/nixpacks/pull/856

PR #863, opened 2023-04-24
Add `--copies` flag to python venv

heojay:
I've noticed that when building a Docker image for Python using Nixpacks, the current implementation relies on `venv` for the virtual environment setup, which by default uses symbolic links to `/root/.nix-profile/bin/python` for Python installation. However, this can cause permission issues when running the image as a non-root user.

To address this issue, I propose enabling the `--copies` flag for `venv`, which will create a copy of the Python executable instead of using symbolic links. This should resolve the permission issues and allow the image to be run as a non-root user.

### Related docs

- https://docs.python.org/3/library/venv.html#creating-virtual-environments


coffee-cup (approved):
Thanks!

PR #881, opened 2023-05-11
Add prettier for doc

Miloas:
I found there are many unused imports in doc source code, so I tried to add prettier. 

And I think we need `pre-commit` too, we can use raw `pre-commit` or add a `project.json` at root and then use `husky`. so before we make a decision, just add prettier for now.


Milo123459:
i dont think we need a precommit hook for this. just prettier is fine imo

Miloas:
> i dont think we need a precommit hook for this. just prettier is fine imo

if we dont add rust lint and format I think that's ok, because only railway people may modify the doc source code 🤔️

docs/package.json
```diff
@@ -5,7 +5,8 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
-    "clean": "rm -rf .next"
+    "clean": "rm -rf .next",
+    "format": "prettier --write \"**/*.{js,ts,tsx,md}\""
```
FinnDore, line 9: Format:check script could be added for linting

FinnDore:
We should add a prettier config as the rest of the repo uses 4 spaces so we should stay consistent here .

Agree with Milo here, pre commit hooks are nasty, we should check files are formatted correctly in ci instead.

docs/package.json
Miloas: done

Miloas:
> 

Are you sure 4 spaces? because all the doc source code is 2 spaces based..

Miloas:
> We should add a prettier config as the rest of the repo uses 4 spaces so we should stay consistent here .

Are you sure 4 spaces? because all the doc source code is 2 spaces based..

FinnDore:
> > We should add a prettier config as the rest of the repo uses 4 spaces so we should stay consistent here .
> 
> Are you sure 4 spaces? because all the doc source code is 2 spaces based..

was referring to the rust side of the codebase. But if the doc's are already 2 i guess its fine? will see what milo and jr thinks

Miloas:
cc @Milo123459 @coffee-cup

Miloas:
BTW, if we want indentation consistent, we also can add `.editorconfig`. 
IMO, 2 spaces is fine. (actually I love 2 spaces, lol)

coffee-cup (approved):
Agreed that I don't think we need a pre-commit step.
Thanks for adding prettier! 🎉

coffee-cup (approved):
Agreed that I don't think we need a pre-commit step.
Thanks for adding prettier! 🎉

coffee-cup (approved):
Agreed that I don't think we need a pre-commit step.
Thanks for adding prettier! 🎉

PR #884, opened 2023-05-11
Fix: correct node version parse

### LINKED ISSUE #858: Node provider fails to detect caret-prefixed versions

### Describe the bug

In the Node provider, the primary check for Node versions happen in [src/providers/node/mod.rs#L328](https://github.com/railwayapp/nixpacks/blob/947a417d91cd00b872b4655fc39f595090a6c2d0/src/providers/node/mod.rs#L328). The regex does not capture preceding carets (`^`) - it starts with a digit until a delimiter (`.` etc.) causing caret-prefixed `^` version strings to be ignored entirely.

When the detection above fails, it falls through to [src/providers/node/mod.rs#L334](https://github.com/railwayapp/nixpacks/blob/947a417d91cd00b872b4655fc39f595090a6c2d0/src/providers/node/mod.rs#L334) which also fails caret checking (it only checks for GTE versions i.e. `>=`).

Caret versions should be supported because they are valid (intended for bumping versions to `minor.patch` versions (`^18` => `18.$latestMinor.$latestPatch`) and widely-used.

Relevant block of code:

https://github.com/railwayapp/nixpacks/blob/947a417d91cd00b872b4655fc39f595090a6c2d0/src/providers/node/mod.rs#L315-L339

---

As a side note, it looks like we're returning the default specified Node version (v16, as a const in [src/providers/node/mod.rs#L25](https://github.com/railwayapp/nixpacks/blob/947a417d91cd00b872b4655fc39f595090a6c2d0/src/providers/node/mod.rs#L25)) regardless of whatever `NODE_VERSION` is provided (in env, `package.json`, etc.) This looks like it should be returning the `node_version` from [src/providers/node/mod.rs#L317-L320](https://github.com/railwayapp/nixpacks/blob/947a417d91cd00b872b4655fc39f595090a6c2d0/src/providers/node/mod.rs#L317-L320) (where a `match` is performed on provided Node version, defaulting to the v16 const) instead of the default version const itselt, but I suppose that's intended because it may result in unintended behaviour when a non-supported version string is provided.

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

n/a

Miloas:
This PR:
fix #858 

#858 describe our current implementation cant parse `^14.10.3` correctly, it will return the default node version, but it should return `14`.

In real world `node-semver` is more complicated. You can check how npm handle this https://github.com/npm/node-semver

Fortunately, we have a great crate can do same thing https://github.com/felipesere/node-semver-rs

## How I think the original implementation?
**inputs (one of):** 
1. package.json -> engines -> node (node semver)
2. .nvmrc (may prefixed `v`)
3. NODE_VERSION (node semver)

**outputs:**
node version that statisfied the inputs version (one of):
1. 14
2. 16
3. 18

### DIFF
--- a/src/providers/node/mod.rs
+++ b/src/providers/node/mod.rs
@@ -13,7 +13,6 @@ use crate::nixpacks::{
 use anyhow::Result;
 use node_semver::Range;
 use path_slash::PathExt;
-use regex::Regex;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::{HashMap, HashSet};
@@ -23,7 +22,7 @@ mod turborepo;
 
 pub const NODE_OVERLAY: &str = "https://github.com/railwayapp/nix-npm-overlay/archive/main.tar.gz";
 
-const DEFAULT_NODE_PKG_NAME: &str = "nodejs-16_x";
+const DEFAULT_NODE_VERSION: u32 = 16;
 const AVAILABLE_NODE_VERSIONS: &[u32] = &[14, 16, 18];
 
 const YARN_CACHE_DIR: &str = "/usr/local/share/.cache/yarn/v6";
@@ -299,6 +298,7 @@ impl NodeProvider {
         app: &App,
         environment: &Environment,
     ) -> Result<Pkg> {
+        let default_node_pkg_name = version_number_to_pkg(DEFAULT_NODE_VERSION);
         let env_node_version = environment.get_config_variable("NODE_VERSION");
 
         let pkg_node_version = package_json
@@ -317,12 +317,12 @@ impl NodeProvider {
 
         let node_version = match node_version {
             Some(node_version) => node_version,
-            None => return Ok(Pkg::new(DEFAULT_NODE_PKG_NAME)),
+            None => return Ok(Pkg::new(default_node_pkg_name.as_str())),
         };
 
         // Any version will work, use latest
         if node_version == "*" {
-            return Ok(Pkg::new(DEFAULT_NODE_PKG_NAME));
+            return Ok(Pkg::new(default_node_pkg_name.as_str()));
         }
 
         let node_pkg = parse_node_version_into_pkg(&node_version);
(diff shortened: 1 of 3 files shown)


Miloas:
@half0wl please would you take a look this pr?

Miloas:
BTW, I found every languages may have their own semver rules (for example, rust semver is different of node semver, most of languages use standard semver), so I cant port it to other languages. But I think other provider may have same issue.

src/providers/node/mod.rs
```diff
@@ -580,16 +571,19 @@ fn version_number_to_pkg(version: u32) -> String {
...
-    if let Some(captures) = matches.get(0) {
-        match captures[1].parse::<u32>() {
-            Ok(version) => return Some(version_number_to_pkg(version)),
-            Err(_e) => {}
+fn parse_node_version_into_pkg(node_version: &str) -> String {
+    let range: Range = node_version.parse().unwrap_or_else(|_| {
+        println!("Warning: node version {node_version} is not valid, using default node version {DEFAULT_NODE_PKG_NAME}");
+        Range::parse(DEFAULT_NODE_VERSION.to_string()).unwrap()
+    });
+    for version_number in AVAILABLE_NODE_VERSIONS {
```
coffee-cup, line 579: Should this start on the newest node version first so that `>=16` resolves to `18` instead of `16`

src/providers/node/mod.rs
```diff
@@ -22,7 +23,8 @@ mod turborepo;
 
 pub const NODE_OVERLAY: &str = "https://github.com/railwayapp/nix-npm-overlay/archive/main.tar.gz";
 
-const DEFAULT_NODE_PKG_NAME: &str = "nodejs-16_x";
+const DEFAULT_NODE_VERSION: u32 = 16;
+const DEFAULT_NODE_PKG_NAME: &str = formatcp!("nodejs-{DEFAULT_NODE_VERSION}_x");
```
coffee-cup, line 27: I don't think we need to have the default node package in String format here, which would reduce the need for the `const_format` dependency. We can instead just convert the `DEFAULT_NODE_VERSION` to a string when we resolve to a pacakge

Miloas: @coffee-cup Cant easily resolve `DEFAULT_NODE_VERSION` to string in tests. So I add a new method to get node version number from pkg name `pkg_to_version_number`, inputs like `nodejs-16_x` outputs will be `16`

src/providers/node/mod.rs
Miloas: done

src/providers/node/mod.rs
Miloas: after ^, we can keep `DEFAULT_NODE_PKG_NAME` and dont need `formatcp`

Miloas:
Rust clippy: "you looks not smart" 😢

src/providers/node/mod.rs
```diff
@@ -580,16 +570,34 @@ fn version_number_to_pkg(version: u32) -> String {
...
+fn pkg_to_version_number(pkg: &str) -> Option<u32> {
+    let re = Regex::new(r"nodejs-(?P<version>\d+)_x").unwrap();
+    match re.captures(pkg) {
+        None => None,
+        Some(captures) => {
+            let version_number: u32 = captures["version"].parse().unwrap();
+            Some(version_number)
         }
     }
+}
```
coffee-cup, line 582: This function only really exists to parse `DEFAULT_NODE_PKG_NAME`. Instead, I think we should have the default node version be a number and then simply call `version_number_to_pkg` on it when we need to return a package string.

Miloas: ok done

half0wl (approved):
🥳 

> BTW, I found every languages may have their own semver rules (for example, rust semver is different of node semver, most of languages use standard semver), so I cant port it to other languages. But I think other provider may have same issue.

Yeah, good call pulling in `node-semver` to deal with this! Happy to see the regex gone ;-)

PR #887, opened 2023-05-12
Add tracing logging

Miloas:
This PR tried to add a logging library for better user and developer experience. 
Only transformed all our `println!` with logging library.

Why [tracing](https://github.com/tokio-rs/tracing)? 
Rust compiler using it (https://github.com/rust-lang/rust/pull/74726). 
It gives us structured logging power, then we(also other service using nixpacks) can easily collecting and parse building logs. (expecially error logs)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,7 +63,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
 dependencies = [
  "quote",
- "syn",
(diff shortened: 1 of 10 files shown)


coffee-cup:
I'm not sure I agree with this PR change. I can see using a tracing library for warnings and errors. But for main logging it really changes what the output looks like

**Now**

image

**Previous**

image

coffee-cup:
Going to close this for the reasons listed above. I'm happy to move to a better more standardized logging system. But the quality of logs needs to go up, not down. The most important logs is the build plan at the top which shows how the app will be built.

PR #890, opened 2023-05-13
Try fix doc ci

Miloas:
This PR...


Miloas:
I am sorry my mistake block the work of @pinkforest, and need @coffee-cup help me fix. I think ci works fine now 🙏

pinkforest:
Could we also get rid of the warnings about default exports ?

Miloas:
> Could we also get rid of the warnings about default exports ?

I think this is a upstream issue (https://github.com/markdoc/next.js/pull/15), after this pr merged, I will submit a new pr to fix

PR #679, opened 2022-11-16
migrate to clap derive api

nebulatgs:
- Migrates to the Clap Derive API instead of the Builder API.
- Upgrades to Clap v4

PR #905, opened 2023-05-26
Use list-like flags for flags that take a list of options

josegonzalez:
This PR migrates the existing flags that use a list of values for a single flag (`--flag value1 value2`) to a more common format of list of flags (`--flag value1 --flag value2`). This allows folks to more seamlessly switch from other CLI tools like `docker` or `pack`. 

### DIFF
--- a/docs/pages/docs/cli.md
+++ b/docs/pages/docs/cli.md
@@ -24,25 +24,25 @@ nixpacks build --help
 
 ### Options
 
-|                             |                                                                             |
-| :-------------------------- | :-------------------------------------------------------------------------- |
-| `--install-cmd <cmd>`, `-i` | Specify the install command                                                 |
-| `--build-cmd <cmd>`, `-b`   | Specify the build command                                                   |
-| `--start-cmd <cmd>`, `-s`   | Specify the start command                                                   |
-| `--name <name>`             | Name for the built image                                                    |
-| `--env <envs...>`           | Provide environment variables to your build.                                |
-| `--pkgs <pkgs...>`, `-p`    | Provide additional Nix packages to install in the environment               |
-| `--apt <pkgs...>`           | Provide additional apt packages to install in the environment               |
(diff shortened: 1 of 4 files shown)


coffee-cup:
Going to hold off on merging this PR for now since it requires a major version update

josegonzalez:
Yeah it doesn't. I was thinking that maybe we could introduce parsing of each option to see if there should be splitting but idk if there is a shell word-splitting library in rust we can pull in.

coffee-cup:
Honestly, I believe we originally had that form of list arg parsing initially and migrated it to what we have currently in this PR https://github.com/railwayapp/nixpacks/pull/679. I will discuss with the team the best option moving forward. Unfortunately, we are unlikely to make a major version update just for this change.

josegonzalez:
If we did shell-parsing of the values when there is only one value, would that be enough to get this into a minor?

josegonzalez:
One thing I noticed is that at least the `pkgs` flag seems to act as I desire with the latest stable release. Let me do a bit more testing to see if this PR is necessary (and when it is necessary, if at all).

josegonzalez:
@coffee-cup is there anything I can do to help make this change more palatable?

josegonzalez:
Is there any interest in this PR at all? If not, I can just close the PR.

josegonzalez:
Closing as there hasn't been any interest from upstream in reviewing/merging the PR.

coffee-cup:
Hi @josegonzalez,
Not that there hasn't been any interest in merging this. It is just a major breaking change so was planning on waiting to merge it with some other breaking changes in the future (no release date). I will revisit this in the future when those changes are made.

josegonzalez:
Just updating, Not sure what happened but seems like list-like flags (for labels at least) suddenly work in main (the latest nixpacks cli does locally anyways)! Maybe someone jingled some keys :)

Issue #826, opened 2023-03-09
nixpacks relies on outdated version of `deno` and other packages


ifiokjr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Attempting to use an up to date version of `Deno` results in an error. 

I'm trying to use [this feature released in January](https://deno.com/blog/v1.30#denojson-becomes-an-import-map) which allows the `deno.jsonc` file to be used in lieu of an `import_map.json`.

Unfortunately the version of deno is not supported wiva nix pkgs. 

To fix this, the version of `nixpkgs` being used will need to be updated.

https://github.com/railwayapp/nixpacks/blob/ec545bbd725d6c2a6cab9031d22e60bc798178f3/src/nixpacks/nix/mod.rs#L8-L11

I'm not sure how feasible it is. 

### To reproduce

Running the following file structure will fail with the following error.

```markup
error: Relative import path "~" not prefixed with / or ./ or ../
```

##### `deno.jsonc`

```jsonc
{
  "imports": {
    "~/": "./"
  }
}
```

##### `mod.ts`

```ts
import { print } from '~/local.ts';
print();
```

##### `local.ts`

```ts
export function print() {
  console.log("i PriNt gOoD");
}
```

##### `nixpacks.toml`

```toml
[phases.setup]
nixPkgs = ["...", "deno"]

[start]
cmd = "deno run mod.ts"
```

### Environment

This fails while using nixpacks on `railway.app`.

Locally, I'm running macos Ventura 13.2.1, and nixpacks 1.4.1.

coffee-cup:
Yes this is known. We are looking into ways of gracfully updating the Nixpkg archive without breaking existing builds that rely on the current default versions.

In the meantime, you can update the archive version specific to your app https://nixpacks.com/docs/configuration/file#nixpkgs-archive

Something like

```toml
[phases.setup]
nixPkgs = ["...", "deno"]
nixpkgsArchive = "293a28df6d7ff3dec1e61e37cc4ee6e6c0fb0847"

[start]
cmd = "deno run mod.ts"
```

ifiokjr:
Oh, I've just seen this in your documentation https://nixpacks.com/docs/configuration/file#nixpkgs-archive

Adding this line fixed it for me:

```diff
[phases.setup] 
+ nixpkgsArchive = 'a71e45961e88c8a6dde6287fa1e061f30f8c2fb7'
nixPkgs = ["...", "deno"]
```

PR #912, opened 2023-06-17
chore: update nixpacks and add go 1.19/1.20 support

stillmatic:
This PR updates the nixpacks commit hash and updates the Go provider with 1.19 and 1.20.

I followed https://github.com/railwayapp/nixpacks/pull/747 to update the commit hash. It seems like the bot is broken and hasn't updated in 6 months. 

I updated the custom version in the custom go version example and confirmed that it built correctly.

### DIFF
--- a/docs/pages/docs/providers/go.md
+++ b/docs/pages/docs/providers/go.md
@@ -16,6 +16,8 @@ The following Go version are available
 
 - `1.17` (Default)
 - `1.18`
+- `1.19`
+- `1.20`
 
 The version is parsed from the `go.mod` file.
 
--- a/examples/go-custom-version/go.mod
+++ b/examples/go-custom-version/go.mod
@@ -1,3 +1,3 @@
 module versiontest
 
-go 1.18
+go 1.20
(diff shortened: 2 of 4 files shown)


stillmatic:
@coffee-cup  is https://github.com/railwayapp/nixpacks/issues/826#issuecomment-1461288199 still the preferred way to handle this?

coffee-cup:
> @coffee-cup is [#826 (comment)](https://github.com/railwayapp/nixpacks/issues/826#issuecomment-1461288199) still the preferred way to handle this?

Yes that is still preferred. The bot itself isn't broken, it is the tests that are broken. For example, updating the Nix archive updates the default Java/clojure versions, which break how Nixpacks is currently building clojure apps. This global nix archive version isn't scalable and we are thinking of ways of gracefully handling it.

For now, I would recommend to use a specific archive version for Go the newer Go versions (similar to https://github.com/railwayapp/nixpacks/blob/main/src/providers/swift.rs#L17).

stillmatic:
Sounds good. I'm actually just going to use a Dockerfile then.

PR #913, opened 2023-06-18
Fix yarn workspace selection with turborepo

codekrafter:
Yarn uses a different workspace selection syntax than npm, and using turporepo with a yarn project will currently fail if it has to execute a command in a specific workspace.


coffee-cup (approved):
Thanks!

PR #915, opened 2023-06-20
Document Python detection via Pipfile

proog:
This PR updates the docs to mention that Python is detected when a `Pipfile` is found. Pipfile behavior is already documented for setup but is absent from the introduction.


coffee-cup (approved):
Thanks 😄

PR #916, opened 2023-06-22
ref(deno): Add support for deno.jsonc and index.{tsx,jsx}

kamilogorek:
This PR adds support for `deno.jsonc` to be used as the entry point for `deno task`, as it was already used for detecting whether a project is Deno in the first place.  And it does the same for running the default `index` file, by adding `jsx` and `tsx` to the list of checked extensions.

PR Checklist

### DIFF
--- a/docs/pages/docs/providers/deno.md
+++ b/docs/pages/docs/providers/deno.md
@@ -4,7 +4,7 @@ title: Deno
 
 # {% $markdoc.frontmatter.title %}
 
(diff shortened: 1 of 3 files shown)


coffee-cup (approved):
🎉

coffee-cup:
It looks like some of the tests are failing because we are now generating a build step for deno in an example app where we previously were not. Can you please update these snapshots https://github.com/railwayapp/nixpacks/blob/main/CONTRIBUTING.md#snapshot-tests

kamilogorek:
Uh, my bad! Somehow completely missed the fact that snapshot tests exist in that repo. Updated and verified with the CI flow.

PR #917, opened 2023-06-26
Rails: detect usage of sprockets/propshaft in lockfile, not just gemfile

smudge:
In [Rails 6.1](https://rubygems.org/gems/rails/versions/6.1.7.3) and below, sprockets is an implicit dependency and may not be listed in the `Gemfile` (only in the `Gemfile.lock` lockfile).

This means that Railway fails to run `rake assets:precompile` when it otherwise should.

This change also allows for apps which include _other_ implicit dependencies on sprockets or propshaft, which would fix cases where someone pulls in asset pipeline assets via a [Rails engine](https://guides.rubyonrails.org/engines.html), which is common in Admin gems (e.g. `activeadmin`).

### DIFF
--- a/src/providers/ruby.rs
+++ b/src/providers/ruby.rs
@@ -296,7 +296,6 @@ impl RubyProvider {
 
     fn uses_asset_pipeline(&self, app: &App) -> Result<bool> {
         if app.includes_file("Gemfile") {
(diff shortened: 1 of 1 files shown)


coffee-cup (approved):
Thanks!

coffee-cup:
Please update the PR to remove the unused variable and pass linting

PR #929, opened 2023-07-15
Improvements for PHP APIs & routers

aleksrutins:
This PR adds the `NIXPACKS_PHP_FALLBACK_PATH` variable, which allows you to set a fallback to a certain file (like `index.php`).


docs/pages/docs/providers/php.md
```diff
@@ -10,7 +10,8 @@ Note that Laravel apps need an `APP_KEY` environment variable in order to work.
 
 If an `nginx.conf` or `nginx.template.conf` (see [this file](https://github.com/railwayapp/nixpacks/blob/main/src/providers/php/nginx.template.conf) for an example of template syntax) file is found in the project root directory, that configuration will be used.
 
-If a `NIXPACKS_PHP_ROOT_DIR` variable is passed, that will be used as the server root.
+If an `NIXPACKS_PHP_ROOT_DIR` variable is passed, that will be used as the server root.
+If an `NIXPACKS_PHP_FALLBACK_PATH` variable is passed, that will be used as a fallback for the server - for instance, if your app uses `index.php` as a router, you would set this variable to `/index.php`.
```
coffee-cup, line 14:
```suggestion
If a `NIXPACKS_PHP_ROOT_DIR` variable is passed, that will be used as the server root.
If a `NIXPACKS_PHP_FALLBACK_PATH` variable is passed, that will be used as a fallback for the server - for instance, if your app uses `index.php` as a router, you would set this variable to `/index.php`.
```

coffee-cup (approved):

docs/pages/docs/providers/php.md
aleksrutins: I don't know what I was thinking there, thanks.

PR #942, opened 2023-08-07
Support gradle 8

### LINKED ISSUE #921: Gradle 8 support
### Feature request

Hello, could you please add support for gradle 8? I am getting following error

```
Nixpacks build failed
﻿ 
﻿Error: Unsupported Gradle version: 8
```

### Motivation

I have Android + Backend mono repo and I cannot upgrade to new gradle versions because my backend builds start failing if I do

### Contribution

_No response_

heojay:
Resolves #921

- Since gradle nixpkg supports the gradle 8 version, I think we can do the same with nixpacks.
  - https://search.nixos.org/packages?channel=23.05&show=gradle&from=0&size=50&sort=relevance&type=packages&query=gradle

Screen Shot 2023-08-07 at 15 46 00


coffee-cup:
I suspect that this doesn't work correctly because we are on an older version of the Nixpacks archive. Can you please add a test to check if Gradle 8 is correctly used.

heojay:
@coffee-cup  Ok, I will add the test using gradle 8.

heojay:
I don't know much about both zig and gyro, but the reason why the zig-gyro test fails in CI seems to be due to the inaccessibility of the `astrolabe.pm` specified in the file below.

https://github.com/railwayapp/nixpacks/blob/b0b8cd4874ca75ab1f512d92e5f4503f21a0c611/examples/zig-gyro/gyro.lock#L1

Miloas:
I guess thats because `zig 0.11` break the `gyro`.

coffee-cup:
The version of gradle that is actually being installed is Gradle 7 since the Nix archive version is not up to date. 

image


If possible, can you do something like this https://github.com/railwayapp/nixpacks/blob/main/src/providers/elixir.rs#L14 to manually set the archive version to one with Gradle 8.

heojay:
@coffee-cup Sorry, but could you please check this PR? Thank you.

coffee-cup (approved):
Looks good! 🎉

PR #970, opened 2023-09-19
update global nixpkgs archive

coffee-cup:


src/nixpacks/nix/mod.rs
```diff
@@ -8,7 +8,7 @@ pub mod pkg;
 // This line is automatically updated.
 // Last Modified: 2023-01-02 17:04:24 UTC+0000
```
Milo123459, line 9: should this maybe be updated?

src/nixpacks/nix/mod.rs
```diff
@@ -8,7 +8,7 @@ pub mod pkg;
...
 // https://github.com/NixOS/nixpkgs/commit/293a28df6d7ff3dec1e61e37cc4ee6e6c0fb0847
```
Milo123459, line 10: and this

Milo123459 (changes requested):

src/nixpacks/nix/mod.rs
coffee-cup: Yes this is a draft and not done yet

coffee-cup: All updated now 😄

Issue #979, opened 2023-09-25
Springboot application fails to start with default configuration from version >= 2.5


fattila16:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

With the release of `spring 2.5`, if one uses it along with `gradle`, by default it will generate a library next to your executable jar, when you run the `./gradlew build` command. The default start command generated by the `JavaProvider` in nixpacks is `java $JAVA_OPTS -jar -Dserver.port=$PORT build/libs/*.jar`. If there are multiple `*.jar` files in the `build/libs` folder it can point to the one which is not an executable, in this case the `*plain.jar`.

When you run the image you'll get the following error:
```
no main manifest attribute
```

This is because it tries to execute a jar, which is in reality a library.

 Of course the user can modify the `gradle.build` configuration, to not generate the `*plain.jar`, however to provide a better user experience that works without requiring user intervention would be a good thing. (Mentioned here: https://github.com/railwayapp/nixpacks/issues/677)

**Solution**:

From spring v2, there is a dedicated gradle task, called `bootJar` that only generates an executable jar file. If one modifies the `JavaProvider` to do this, for gradle + spring applications the problem goes away. However, this means that spring v1 applications won't work anymore.

I can provide a PR, along with the modifications to implement this.

My question is, is it okay to drop the support of spring v1 applications? The last release was back in 2007. I am not sure how many users of nixpacks use spring v1, so that would be an interesting metric also to decide this.

### To reproduce

Steps to reproduce the bug:

1. Clone an example spring v3 project, for eg.: https://github.com/appuio/example-spring-boot-helloworld
2. Run `nixpacks build .` in the project root
3. Use docker to run the image
4. Results in an error: `no main manifest attribute, in build/libs/app-0.1.1-SNAPSHOT-plain.jar`

### Expected behavior

The expected behaviour  would be that a default spring >= v2.5 project works out of the box without the need to overwrite default configurations.

### Environment

* OS version: macOS 13.5.2 
* nixpacks version: 1.16.0

fattila16:
PR to fix the issue: https://github.com/railwayapp/nixpacks/pull/980

PR #980, opened 2023-09-25
fix/#979 fix springboot running plain jar instead of executable jar

fattila16:
This PR solves the following issue: https://github.com/railwayapp/nixpacks/issues/979

~~**Disclaimer**:  This drops support of any spring version that does not have a `bootJar` task, mainly spring v1~~

* ~~Adjust the `JavaProvider` so that it uses `bootJar` to build an executable JAR instead of `build` which generates a `*plain.jar` also when spring and gradle is detected~~
* ~~Add e2e test cases for spring v2 and v3, also remove spring v1~~
* ~~Update the docs of the `JavaProvider` so that it has information regarding the spring version support, and the build commands~~

As per the comments below PR implements a different solution now:

* Adjust the `JavaProvider` so that if gradle and spring is detected then the start cmd includes a one line to find the executable jar and blacklists the plain.jar
* Add e2e test cases for spring v1, v2 and v3
* Update the docs of the `JavaProvider` so that it includes description for the new start cmd if gradle and spring is detected

- [x] Unit tests are added
- [x] E2E tests are added
- [x] Docs are updated


src/providers/java.rs
```diff
@@ -95,11 +96,20 @@ impl JavaProvider {
         }
     }
 
+    fn get_gradle_build_cmd(&self, build_gradle_content: &String, gradle_exe: &String) -> String {
+        if self.is_using_spring_boot(build_gradle_content) {
```
coffee-cup, line 100: Is it possible to detect if Spring v1 is used and only use `bootJar` if it is version 2+?

fattila16: One can read it from the `build.gradle` file, however I would not go that way. There can be edge cases, where a gradle plugin update or a spring version update changes the versioning scheme. Also one can define it in kotlin and groovy. That makes capturing this a little bit finnicky and I would not want to rely on a bunch of regexp expressions.

If `nixpacks` would like to support spring v1 as well, there is an alternative solution to what I've provided in this PR. Which is the following:

* The problem is that a `*plain.jar` is being generated along with an executable `.jar`
* The start command can point to the `*plain.jar` instead of the executable
* As a solution one can adjust the start command with the following one-liner: 
`java $JAVA_OPTS -jar -Dserver.port=$PORT $(ls -1 build/libs/*jar | grep -v plain)`
* So basically instead of configuring gradle to create only an executable jar, we use `ls` and `grep` to find the executable `jar` instead of the `plain` one

What do you think about this approach?

coffee-cup: I think that solution works! And yeah, supporting v1 isn't 100% necessary, but backwards compatibility is always nice, especially if the solution is as simple as greping for the executable jar

fattila16: I've updated the PR with the new solution

fattila16: Is it possible for me to get a review here? And then I can check whether all the pipelines are green or not. Thank you!

coffee-cup (approved):

coffee-cup:
Thank you 🎉

PR #983, opened 2023-09-27
Update sidebar.ts

9trocode:
This PR add pipeops as part of platforms that support nixpack


coffee-cup (approved):
Ah my bad. I meant to add it earlier. Thank you 😄

Issue #993, opened 2023-10-10
Nixpacks Bug Report


gwbischof:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

`nixpack build .`

gives the following error:
```Error: This manifest requires workspace inheritance, but `inherit_workspace` hasn't been called yet```

when using:
```
[dependencies]
http = { workspace = true }
```
like described in this doc:
https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

### To reproduce

Use the code from this PR:
https://github.com/railwayapp/nixpacks/pull/992

```cd nixpacks/examples/rust-cargo-workspaces```
```nixpacks build .``` 

### Expected behavior

build successful

### Environment

NAME="Linux Mint"
VERSION="21.2 (Victoria)"

```nixpacks --version
nixpacks 1.17.0```

gwbischof:
Maybe this can be fixed by bumping the rust version of nixpacks

PR #995, opened 2023-10-12
Bump rust version

gwbischof:
Bump rust to the latest version so that newer rust workspace features are supported.
See #993


coffee-cup:
Can you please add a test for the workspace problem so we can confirm if it is working

PR #1012, opened 2023-11-20
Fix Node 14 usage

coffee-cup:
Only use the latest Nix archive version for Node != 14


JakeCooper (commented):
Do we want to support Node14? It's EOL'd 3 months ago

I'd be okay just dropping it inline with the upstream support SLO

EDIT: Discussion https://discord.com/channels/713503345364697088/1176264378992050237/1176264535804498031

PR #1014, opened 2023-11-21
add openssl lib when using prisma

coffee-cup:
Fix prisma errors related to missing libssl library


Milo123459:
isnt this redundant as we already add the nix package?

PR #1019, opened 2023-12-03
Update Go versions

aleksrutins:
This adds support for Go 1.19, 1.20, and 1.21, and drops support for 1.17, which is no longer in nixpkgs.


Milo123459:
can we make this better by not hardcoding a list of versions we can use?

PR #1020, opened 2023-12-04
Add support for Java 20/21 in Java & Scala Providers

marcospereira:
## What?

This PR adds Java 20 and 21 support in Java and Scala providers.

Specifically for Scala, it also updates the Eclipse Temurin images to the latest
versions.

- [x] Tests are added/updated if needed
- [x] Docs are updated if needed

PR #1021, opened 2023-12-04
Make Gradle 8 the default version

marcospereira:
## What?

This PR makes Gradle 8 the default version. The first release for Gradle 8 was made on Feb 13, 2023, so it should be safe to make it default (coming from version 7).

### DIFF
--- a/docs/pages/docs/providers/java.md
+++ b/docs/pages/docs/providers/java.md
@@ -25,8 +25,8 @@ The version can be overridden by setting the `NIXPACKS_JDK_VERSION` environment 
 
 The following major Gradle versions are available
 
-- `8`
-- `7` (Default)
+- `8` (Default)
+- `7`
 - `6`
 - `5`
 - `4`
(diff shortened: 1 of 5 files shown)


marcospereira:
@aleksrutins, somewhat.

But I think it is better to have these changes in different pull requests, though. As far as I know, supporting Java 21 does not require making Gradle 8 the default.

aleksrutins:
Just supporting Java 21 doesn't necessarily; but, since it's the new LTS version, #1081 makes it the default, which [does require changing the default Gradle to 8](https://docs.gradle.org/current/userguide/compatibility.html).

marcospereira:
This is not blocked by #1020, but it would be nice to get that done first.

aleksrutins:
not stale

marcospereira:
@coffee-cup @aleksrutins, reviving this. :-)

This is good to go (I've solved a few linting/formatting issues unrelated to my original changes).

coffee-cup (approved):
Looks good!

PR #1026, opened 2023-12-12
feat: add asdf support to python + poetry

iloveitaly:
This PR is not complete! I wanted to get your thoughts on it before putting anymore work into it.

The idea here is to support asdf/.tool-versions for specifying python + poetry versions. This could then
be extended to support node, go, etc in the future but my initial use case is scoped to python.

Interest in this feature being added? If so I can finish implementation + add tests.


iloveitaly:
bump

PR #1048, opened 2024-01-18
update the docker-host and docker-tls-verify flag description for the help command

kcq:
A simple description update for the docker-host and docker-tls-verify flags. No tests are necessary because there are no business logic changes. The docs don't need to be updated because they contain the correct information.


kcq:
The "Is Org Member" check appears to fail in a non-standard way with a code exception... complaining that it's missing input...

coffee-cup (approved):
Thank you!

PR #1052, opened 2024-01-26
fix: automatically add gcc-unwrapped when installing sharp

iloveitaly:
This PR adds the necessary lib to eliminate this error when using the sharp npm package:

```
error: Could not load the "sharp" module using the linux-arm64 runtime
undefined: libstdc++.so.6: cannot open shared object file: No such file or directory
Possible solutions:
```


iloveitaly:
Friendly reminder here—would love to get this merged in if possible.

coffee-cup (approved):
Amazing thank you!

Issue #1030, opened 2023-12-14
Unable to get prisma 5 with nixpacks openssl


Maxim-Filimonov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Following up on [the issue](https://github.com/railwayapp/nixpacks/issues/574) Prisma does support open ssl 3.0.x now.
However, for some reason I'm still not able to use nixpack with Prisma.
I tried installing `libssl-dev` in setup stage and got the same error. 

### To reproduce

After building i get the following error on trying to instantiate `PrismaClient()`
```
PrismaClientInitializationError: Unable to require(`/app/node_modules/.prisma/client/libquery_engine-debian-openssl-3.0.x.so.node`).

Prisma cannot find the required `libssl` system library in your system. Please install openssl-3.0.x and try again.

Details: libssl.so.3: cannot open shared object file: No such file or directory

at new e (/app/node_modules/@prisma/client/runtime/library.js:26:1871)
```

### Expected behavior

Able to build and deploy application using Prisma.

### Environment

nixpacks.toml
```
[phases.setup]
aptPkgs = ["..."]
[phases.build]
cmds = ["...", "bunx prisma generate", "bunx prisma migrate deploy"]
```
package.json
```
 "dependencies": {
    "@prisma/client": "5.7.0",
  },
```
**Bun:** 1.0.17

I have no problems running locally.

coffee-cup:
We have an example that builds and runs no problem using Prisma 5 and bun https://github.com/railwayapp/nixpacks/tree/main/examples/node-bun-prisma.

I can check this out further, but there is evidence of both working so I am not sure what is going on with your specific repo. Do you have a sample repo that doesn't work with Bun?

Maxim-Filimonov:
@coffee-cup I tried using linked example and it fails with different error for some reason. 
```


$ bunx prisma migrate deploy

Prisma schema loaded from prisma/schema.prisma

Datasource "db": PostgreSQL database "railway", schema "public" at "roundhouse.proxy.rlwy.net:17360"

Error: P1001: Can't reach database server at `roundhouse.proxy.rlwy.net`:`17360`

Please make sure your database server is running at `roundhouse.proxy.rlwy.net`:`17360`.

error: script "start" exited with code 1 (SIGHUP)

$ bunx prisma migrate deploy
```
I did add shared variable DATABASE_URL.
[...]

PR #1054, opened 2024-01-26
fix: openssl should be set to lib

iloveitaly:
Without this,  prisma can't find the libraries and will error out


baptisteArno:
So that's related to #1030 ? Is there any work around before this is getting merged?

iloveitaly:
@baptisteArno you can manually add openssl as a lib via the `--libs` cli arg

coffee-cup (approved):

coffee-cup:
Looks like some relevant tests are failing. Could you please look into whats going on there. It might be an issue with older versions of prisma?

coffee-cup:
I'd like to revive this PR as I think it is still needed. Some of the tests are failing and can check it out more deeply next week. It would also be a good idea to add a test for prisma 5

PR #1055, opened 2024-01-26
fix: add correct libs if canvas and puppeteer are setup

iloveitaly:
I'm not familiar with canvas, but I don't see any reason why these two libs would be mutually exclusive.


iloveitaly:
Friendly reminder here—would love to get this merged in if possible.

coffee-cup (approved):

Issue #1042, opened 2024-01-10
add asdf support to python + poetry


iloveitaly:
### Feature request

The idea here is to support asdf/.tool-versions for specifying python + poetry versions. This could then
be extended to support node, go, etc in the future but my initial use case is scoped to python.

### Contribution

https://github.com/railwayapp/nixpacks/pull/1026

zuchka:
@iloveitaly sorry about the delays--I'm looking into this...

zuchka:
folks seem to like this proposal. Are you interested in finishing your PR? It looks like it's not quite done...

zuchka:
@iloveitaly also check this out:
https://github.com/railwayapp/nixpacks/issues/1017#issuecomment-1839761192

maybe worth hopping on that thread?

coffee-cup:
Hi @iloveitaly,
I think having a full on asdf provider could be a good idea. Basically if `.tool-versions` file is found at the root, then we install asdf and all packages defined. Would you be interested in creating a PR for this?

The problem with https://github.com/railwayapp/nixpacks/pull/1026 is that Nix does not support specifying the exact version of packages. It only supports the latest major/minor version. So parsing a tool-version file and then passing the version to Nix does not make a lot of sense in that case.

iloveitaly:
@coffee-cup That makes sense—What do you think about just extracting the major/minor version from asdf and passing that down to the Python builder?

coffee-cup:
That works as a good first step!

PR #1062, opened 2024-02-07
feat: poetry + python version from asdf

iloveitaly:
As discussed in https://github.com/railwayapp/nixpacks/issues/1042. This extracts poetry and python versions from .tool-versions and lays the groundwork for supporting more languages.

Details:

- feat: add asdf support to python + poetry
- feat: initial asdf poetry and python extraction

This PR...

- [x] Tests are added/updated if needed. Can add more once reviewed.


src/providers/python.rs
```diff
@@ -301,6 +316,26 @@ impl PythonProvider {
             .map(|m| m.get(2).unwrap().as_str().to_string()))
     }
 
+    fn parse_tool_versions_python_version(file_content: &str) -> Result<Option<String>> {
+        let asdf_versions = parse_tool_versions_content(file_content);
+
+        // the python version can only specify a major.minor version right now, and not a patch version
+        Ok(asdf_versions.get("python").map(|s| {
+            let parts: Vec<&str> = s.split('.').collect();
+            assert_eq!(
```
coffee-cup, line 325: Instead of asserting here, could we instead just use the major/minor versions and warn that specifying the exact patch version is not supported

coffee-cup (commented):
This is coming together! A small comment about not asserting. On top of that could you please add a small example/test case for python with asdf tool version

PR #1063, opened 2024-02-07
feat: expanded default python variables

iloveitaly:
This set of environment variables seems to be the best set of default production settings
for most python projects.

Can add more tests if needed.


coffee-cup (commented):
These look good, but just wondering where these came from. Could you also please update the Python docs to include these variables.

PR #1066, opened 2024-02-09
feat: dump build command when --out is specified

iloveitaly:
the build command contains a lot of values which are generated from the nixpacks
build process. Without these dumped in the --out directory it's hard to debug
and replicate what's going on with a build.

I have *not* implemented tests, formatting, etc. Would like to get feedback before spending more time on this.


coffee-cup (commented):
This is interesting. Is the use case mainly for debugging what Nixpacks is doing? I wonder if we could instead of a verbose mode where the docker command is output in the logs. Would that solve the problem?

PR #1083, opened 2024-04-08
Update `pdm` to latest

patrick91:
Via https://latest.cat/pdm


patrick91:
@coffee-cup do I need to run `cargo insta test` and update the snapshots? 😊

coffee-cup (approved):
Thanks for the PR :)
I've updated the snapshot tests so it should be good to merge once everything passes

PR #1099, opened 2024-05-07
 Update Bun version via newer NixPkgs archive commit hash

benjamint08:
This PR changes one line in /src/providers/node/mod.rs which uses a newer NixPkgs commit hash which uses Bun 1.1.7. Nixpacks currently uses a much older commit hash which uses Bun 1.0 which lacks a LOT of features that 1.1.7 has that are needed for the apps I develop.

I was not able to run tests because I kept getting weird docker errors from Cargo- but Bun works fine:

image

Thanks to anyone that sees this.

### DIFF
--- a/src/providers/node/mod.rs
+++ b/src/providers/node/mod.rs
@@ -23,7 +23,7 @@ mod turborepo;
 
 pub const NODE_OVERLAY: &str = "https://github.com/railwayapp/nix-npm-overlay/archive/main.tar.gz";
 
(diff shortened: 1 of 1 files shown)


coffee-cup:
Thanks for the PR! However, closing in favour of https://github.com/railwayapp/nixpacks/pull/1106 which updates to the latest Bun and fixes a few related things. Once that is released you should have access to the latest Bun on Railway

PR #1115, opened 2024-06-09
feat: dump build command when --out is specified

iloveitaly:
Follow-up from https://github.com/railwayapp/nixpacks/pull/1066

> Is the use case mainly for debugging what Nixpacks is doing? I wonder if we could instead of a verbose mode where the docker command is output in the logs. Would that solve the problem?

Yes. Primarily for debugging.

Without this, it's hard to actually recreate what nixpacks is doing when building a dockerfile to 
iterate on the Dockerfile config locally.

Outputting the build command would be fine, but it seems better to me to persist the necessary commands to build the
entire docker container exactly like nixpacks does when --dump is used.

### DIFF
--- a/src/nixpacks/builder/docker/docker_image_builder.rs
+++ b/src/nixpacks/builder/docker/docker_image_builder.rs
@@ -37,6 +37,11 @@ fn get_output_dir(app_src: &str, options: &DockerBuilderOptions) -> Result<Outpu
     }
 }
 
+fn command_to_string(command: &Command) -> String {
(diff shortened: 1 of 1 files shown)


iloveitaly:
This also fixes a listing error that I believe exists in master 70e0a6ab41f62438191548ec1425cf8046060bef

iloveitaly:
Friendly reminder on this pull request! Feel free to let me know if there are any updates needed here.

iloveitaly:
Friendly reminder about this pull request! Please let me know if there is anything else that needs to be addressed.

coffee-cup (approved):
Looks good! Sorry about such the long delay to get this merged

PR #1116, opened 2024-06-09
feat: asdf support

iloveitaly:
Follow-up on https://github.com/railwayapp/nixpacks/pull/1062

> Instead of asserting here, could we instead just use the major/minor versions and warn that specifying the exact patch version is not supported

This is fixed!


iloveitaly:
Friendly reminder on this pull request! Please provide an update or this will be closed soon.

JakeCooper (approved):

PR #1120, opened 2024-06-15
fix available / not available versions

brody192:
Got asked plenty of times what node/python versions are supported / not supported.


brody192:
cc @zuchka

JakeCooper (approved):

PR #1123, opened 2024-06-18
Add Gleam To Sidebar

brianmichel:
It seems that Gleam support was added a while back, but the docs for it were never added to the sidebar so you might not know that nixpacks actually support building Gleam projects.


brianmichel:
@coffee-cup no idea who I should tag for something this small. I checked the contributing documentation and didn't see any specific guidelines for reviewing, please let me know if I missed something...

JakeCooper (approved):

PR #1127, opened 2024-06-25
feat(tests): better cleanup on docker tests

CoderJoshDK:
The `docker_run_tests`, tests, do not properly cleanup after itself. The changes made are as follows:

* Remove generated [anonymous volumes](https://docs.docker.com/reference/cli/docker/container/rm/#volumes):
    - add `-v` to `docker rm`
* Remove generated image:
    - call `docker rmi` after containers all stopped and removed
* Remove all containers related to an image and not just the most recent one
* Cleanup images that complete quickly:
    - call `stop_and_remove_container_by_image` on all images, not just
      ones that hang
* Convert fake async function to normal function:
    - `get_container_ids_from_image` never awaits anything. It was not a
      real async function. As such, the async syntax was removed for it
* Gracefully cleanup failed docker builds
    - some tests generate extra resources. They are now cleaned up and removed, even on failed docker builds.

Some standing issues with the tests:
* Not all tests are able to successfully build on local machine.  
* `test_docker_host` will set a broken host on all tests that follow it. Not an issue if tests are done in isolation. But it becomes an issue if all tests are run together. (at least on my machine)


JakeCooper:
Hey can you resolve conflicts and fix lint?

CoderJoshDK:
The lint error was because the `clippy` command provided in the `CONTRIBUTING.md` file was not the same as the command used by the CI. So, I also updated that command (in the contributing file.)

CoderJoshDK:
Sorry, missed that there was a conflict too. Should be all fixed now.

coffee-cup (approved):
Thanks!

PR #1129, opened 2024-06-28
feat: add a params to specify the docker cert path

cchalop1:
I made this PR to add a params to specify the docker cert path.

Tell me what you thinks about it ?

Do you have any idea to test it ?


coffee-cup:
Hi @cchalop1, Thanks for the PR.

While I can see this as useful, I'm wondering about your use case and if running NIxpacks with the `--out {dir}` command would work for you. This outputs a Dockerfile and all the files that Nixpacks uses to generate a Docker build with. You can use this to run the docker command manually and include any config necessary.

Would this work? Not opposed to adding the arg to Nixpacks. But I worry about how much Docker config Nixpacks should import and simply pass on to Docker.

cchalop1:
Sorry for my late response, I understand your point of view @coffee-cup. However, there is already a variable to specify a Docker host, called docker-host, so it would make sense to have another variable just for establishing a secure connection with this host, wouldn't it?"

PR #1152, opened 2024-08-02
Support Bun 1.1.20

dayblox:
Update Nix archive used for Node so we can support Bun 1.1.20

https://github.com/NixOS/nixpkgs/commit/e05605ec414618eab4a7a6aea8b38f6fbbcc8f08


dayblox:
@coffee-cup I'm unable to add the label, could you? 🙏

coffee-cup (approved):
Thanks for the updated :D

PR #1165, opened 2024-08-25
Feat/docker cert path

cchalop1:
Hi @coffee-cup,

I'm recreating this pull request following the closure of the previous one. As discussed earlier, I'm proposing to add a parameter to specify the Docker cert path.

I understand your concern about not wanting to integrate too much Docker configuration into Nixpacks. However, since there is already a variable to specify a Docker host (docker-host), it seems logical to add another variable specifically for establishing a secure connection with this host. This would make it easier for users who want to secure their connection without manually configuring every aspect.

Please let me know your thoughts on this. I'm also open to suggestions on how to test this new feature.

Thank you for your understanding and time!


coffee-cup (approved):
Thats very understandable and I agree they go hand in hand. Thanks for the PR :D

cchalop1:
Hi @coffee-cup, thanks i appreciate 👍

cchalop1:
I have fix the linter i should be good

PR #1176, opened 2024-09-01
Add removal of `DOCKER_HOST` env when `docker_host` is `None`

guoard:
This PR enhances the handling of the `DOCKER_HOST` environment variable in the codebase.

#### Changes:
- Replaced the existing `if let` structure with a `match` statement.
- Added logic to remove the `DOCKER_HOST` environment variable when the `docker_host` option is `None`.

#### Impact:
Without this change, running tests locally would fail because the `DOCKER_HOST` variable wouldn't be removed when `docker_host` was `None`.


coffee-cup:
Thanks for the PR! I was just in that area so included this update in https://github.com/railwayapp/nixpacks/pull/1180

PR #1177, opened 2024-09-02
Fix fill_auto_in_vec function doc test

guoard:
This PR fixes the documentation test (`doctest`) for the `fill_auto_in_vec` function, which was previously failing when running the command:

```bash
cargo test --doc
```

#### **Changes Made:**
- Updated the doctest to use the `.into()` method for converting `&str` to `String`, ensuring type consistency without cluttering the test code.
- Made the `utils` module public to ensure that the doctest can access the `fill_auto_in_vec` function correctly.

#### **Consideration:**
If these changes seem excessive or if maintaining the doctest is redundant (given that the function is already thoroughly tested in the `test` module within the same file), you might consider removing the doctest altogether to avoid duplicating efforts.

### DIFF
--- a/src/nixpacks/plan/mod.rs
+++ b/src/nixpacks/plan/mod.rs
@@ -18,7 +18,7 @@ pub mod merge;
 pub mod phase;
 pub mod pretty_print;
 mod topological_sort;
-mod utils;
+pub mod utils;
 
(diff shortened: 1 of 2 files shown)


coffee-cup (approved):
thanks!

PR #1178, opened 2024-09-02
fix: automatically add cairo lib dependency when pip package is detected

iloveitaly:
This PR automatically adds the cairo lib when it's detected in the dependency list.


coffee-cup (approved):
🎉

Issue #990, opened 2023-10-09
Support running as non root user


flybayer:
### Feature request

I would like ability to generate dockerfile that will run as a non root user. Required for apps that run headless chrome for example.

janus-reith:
This has been brought up earlier in https://github.com/railwayapp/nixpacks/issues/449

I wonder if this has been addressed in the meantime?

PR #1193, opened 2024-09-22
support running with custom non-root user

### LINKED ISSUE #449: Final start command is run as `root`

### Describe the bug

The final start command is run as `root` in the deployed container.
This is bad practice, and can open up vulnerabilities for the app run as `root`.

This also causes certain programs like `puppeteer` to refuse to launch (`ERROR:zygote_host_impl_linux.cc(90)] Running as root without --no-sandbox is not supported.`)

### To reproduce

Build a project with Nixpacks.
Run the resulting OCI image.
The start command will be run as root.

### Expected behavior

The start command should be deescalated and run as another PID.

### Environment

Windows 11, Nixpacks v0.3.8

(affects all versions)

coffee-cup:
This PR adds support for running the start command as a non-root user. The user is frist created with `useradd`.

This is a very simply implementation of custom user support and if you have a another use case that requires additonal user configuration, please leave a comment or create an issue to open up a discussion.

Closes #449, #990

## Configuration

`user` is added as a field to the start phase and can be configured with a `nixpacks.toml` file like this

```toml
# ...

[start]
user = "myuser"
```

or via the command line by passing custom config. For example

```bash
$ nixpacks build examples/custom-user --name user --json-plan '{ "start": { "user": "myuser2" } }'
```


fattila16:
When it comes to serving static assets, `nginx` is used. With the introduction of the custom non-root user, would this break that?  As `nginx` needs root access to write `/var/log/nginx`, `/var/cache/nginx` and for whatever else I'm not 100% familiar what nginx needs to run.

coffee-cup:
This will not change anything by default so nothing should break when nginx is used unless a non-root user is manually specified.

PR #1195, opened 2024-09-22
Update install steps in getting-started.md

aisk:
The `install.md` already changed to using the brew's main tap to install nixpacks.


coffee-cup (approved):
🎉

PR #1206, opened 2024-10-24
fix: cairo as lib and not package

iloveitaly:
This is a follow-on from #1178. This needs to be specified as a lib otherwise it won't be picked up by python.

The strange thing I'm running into here is Cairo does not show up in the build information dump:

```
╔══════════════════════════════ Nixpacks v1.29.0 ══════════════════════════════╗
║ setup      │ python312, gcc                                                  ║
║──────────────────────────────────────────────────────────────────────────────║
║ install    │ python -m venv --copies /opt/venv && . /opt/venv/bin/activate   ║
║            │ && pip install poetry==$NIXPACKS_POETRY_VERSION && poetry       ║
║            │ install --no-dev --no-interaction --no-ansi                     ║
║──────────────────────────────────────────────────────────────────────────────║
║ start      │ python main.py                                                  ║
╚══════════════════════════════════════════════════════════════════════════════╝
```

the failing lint seems to be unrelated.


coffee-cup (approved):
Thanks!

PR #1207, opened 2024-10-24
fix: better elixir docs and defaults

iloveitaly:
Two main improvements:

- **docs: improve elixir build documentation**
- **fix: better elixir defaults, missing packages**

I don't think additional tests are needed here. Refactored to pattern off of the python provider a bit more closely.


iloveitaly:
Friendly reminder regarding this pull request! Please let me know if there are any further updates needed.

src/providers/elixir.rs
```diff
@@ -51,9 +45,6 @@ impl Provider for ElixirProvider {
             build_phase.add_cmd("mix assets.deploy".to_string());
         }
 
-        if mix_exs_content.contains("postgrex") && mix_exs_content.contains("ecto") {
-            build_phase.add_cmd("mix ecto.setup");
```
coffee-cup, line 55: Why was this removed?

coffee-cup (commented):
Thanks for the PR! Sorry its taken so long to get to and review.

I've updated the snapshot tests for elixir that was failing the build. Just one comment about the ecto.setup being removed, but besides that, it looks good to go.

src/providers/elixir.rs
iloveitaly: Ah, for me I didn't want this. Will add it back in.

The risk with running ecto.setup is it generally requires a DB connection, which is not available when the image is being built.

coffee-cup (approved):

PR #1208, opened 2024-10-24
docs: add link to shell completions

iloveitaly:
docs link to zsh shell completions.

### DIFF
--- a/docs/pages/docs/install.md
+++ b/docs/pages/docs/install.md
@@ -69,3 +69,7 @@ Build and install from source using [Rust](https://www.rust-lang.org/tools/insta
 ```sh
 cargo install nixpacks
 ```
+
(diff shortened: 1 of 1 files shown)


brody192:
Hmmm, I don't think we are too keen about linking out to a repo from a 3rd party.

iloveitaly:
Makes sense. Would you consider bundling this into the nixpacks installation?

iloveitaly:
@coffee-cup do you know where the brew installation logic is? Not sure the best way to package completions into this repo / build system. Point me in the right direction and I'll attempt to make the time for it.

coffee-cup:
@iloveitaly It is in the core Homebrew repo https://github.com/Homebrew/homebrew-core/blob/master/Formula/r/railway.rb

PR #1219, opened 2024-11-13
dokku docs

iloveitaly:
Adding documentation about dokku hosting provider.

### DIFF
--- a/docs/sidebar.ts
+++ b/docs/sidebar.ts
@@ -71,7 +71,7 @@ export const sidebarItems: ISidebarSection[] = [
       { text: "PipeOps", href: "/docs/deploying/pipeops" },
       { text: "Dokploy", href: "/docs/deploying/dokploy" },
       { text: "GitHub Actions", href: "/docs/deploying/github-actions" },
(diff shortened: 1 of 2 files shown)


docs/sidebar.ts
```diff
@@ -71,6 +71,7 @@ export const sidebarItems: ISidebarSection[] = [
       { text: "PipeOps", href: "/docs/deploying/pipeops" },
       { text: "Dokploy", href: "/docs/deploying/dokploy" },
       { text: "GitHub Actions", href: "/docs/deploying/github-actions" },
+      { text: "GitHub Actions", href: "/docs/deploying/dokku" },
```
FarazPatankar, line 74: `text` should probably be `Dokku`?

FarazPatankar (changes requested):

docs/sidebar.ts
iloveitaly:
```suggestion
      { text: "Dokku", href: "/docs/deploying/dokku" },
```

PR #1220, opened 2024-11-13
multi arch docs

iloveitaly:
Docs on support for multi-arch build. this was confusing to me when first coming to the project.


coffee-cup (approved):
ty!

PR #1221, opened 2024-11-13
Update Clojure binary name handling

FarazPatankar:
This PR updates how we look for Clojure binaries to set the start command. Upon merging:
- We will support `.jar` files with our without `-standalone` in the name
- We will ignore `.jar` files with `-SNAPSHOT.jar` in the name

I have also added an example using `clojure-luminus` which prompted this PR.


src/providers/clojure.rs
```diff
@@ -47,7 +47,9 @@ impl Provider for ClojureProvider {
         let mut build = Phase::build(Some(format!("{build_cmd}; {move_file_cmd}")));
         build.depends_on_phase("setup");
 
-        let start = StartPhase::new("JAR_FILE=$(find ./target -name \"*standalone.jar\"); bash -c \"java $JAVA_OPTS -jar $JAR_FILE\"");
+        let start = StartPhase::new(
```
coffee-cup, line 50: Does the above `move_file_cmd` also need to be updated?

FarazPatankar: It shouldn't be but I'll do another pass today and see if I can make both of them consistent.

coffee-cup (approved):

PR #1223, opened 2024-11-13
feat: node 23, more recent patch versions'

iloveitaly:
I needed more recent node patch versions. Updated nix sha and added support for node 23.


coffee-cup (approved):

coffee-cup:
I seem to be getting a prisma unknown binary error with this change when running it locally

image

I don't see this error with the current version of Nixpacks. I'm guessing something to do with updating the node version? But they are still pulling node v18. I'm unsure atm and can do a bit more digging next week.

iloveitaly:
I got this working. There were some really old versions of Prisma and Node that were causing problems with more recent SSL libraries.

coffee-cup (approved):
Thanks for figuring out the ssl issues!

PR #1230, opened 2024-11-28
feat: added support for `lts/*` syntax in nvmrc

### LINKED ISSUE #1190: Nixpacks Node.js Build Fails with .nvmrc File Containing Non-Numeric Version (e.g., lts/iron)

### Describe the bug
When attempting to build a Node.js project with Nixpacks, the build fails if there is an .nvmrc file that contains a version string like `lts/iron`. It seems that the code only recognizes versions in the format `(e.g., 20.13.1)`, but nvmrc files may contain other valid formats such as `lts/iron`.

**Current behaviour:** 
The build fails because the version string in the .nvmrc file is not recognized if its a string (lts/iron) and throws following error:

```js
Syntax error: unexpected token "T_BASIC_UNESCAPED" at line 1 with value ": node version lts/iron is not valid
```

### To reproduce
1. Add an .nvmrc file with the content specifying a version as string like` lts/iron.`
2. Try to build with Nixpacks.

### Environment

Nixpacks: v1.28.1

shubhexists:
Fixes #1190 

PR Checklist 

### DIFF
--- a/src/nixpacks/plan/mod.rs
+++ b/src/nixpacks/plan/mod.rs
@@ -408,6 +408,47 @@ mod test {
     }
 
     #[test]
+    fn test_to_json_and_from_json() {
+        let original_plan = BuildPlan::from_toml(
+            r#"
+            [phases.setup]
+            nixPkgs = ["nodejs", "yarn"]
+            aptPkgs = ["git"]
+
+            [phases.install]
+            cmds = ["yarn install"]
+            cacheDirectories = ["node_modules"]
+            dependsOn = ["setup"]
+
+            [phases.build]
+            cmds = ["yarn build"]
+            dependsOn = ["install"]
+
+            [start]
+            cmd = "yarn start"
+            "#,
+        )
+        .unwrap();
+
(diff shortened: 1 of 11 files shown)


shubhexists:
To help me better understand the codebase, I added a test for the  `to_json` function of the `BuildPlan` struct as they were not already present.
Let me know if I have to remove it as it is out of scope for this PR.

shubhexists:
Hi @brody192, There was no mention for `.nvmrc` in the current docs in `node` provider
I added that 
Let me know if it looks fine : D

brody192:
Awsome, thanks! looks good.

src/nixpacks/plan/mod.rs
```diff
@@ -407,6 +407,47 @@ mod test {
         assert_eq!(result, env_plan);
     }
 
+    #[test]
+    fn test_to_json_and_from_json() {
```
coffee-cup, line 411: good test 🙂

coffee-cup (approved):
Looks great thanks!

PR #1243, opened 2024-12-16
Use latest Deno 2 version

aleksrutins:
This also adds support to `BuildPlan::pin` for pinning the build plan to any custom archive.


aleksrutins:
This is not stale. How can I reopen it?

brody192:
Sorry about that! we'll be looking at this after we are back from winter break!

src/nixpacks/plan/mod.rs
```diff
@@ -46,6 +49,8 @@ pub struct BuildPlan {
 
     pub phases: Option<Phases>,
 
+    pub pinned_archive: Option<String>,
```
coffee-cup, line 52: This is not necessary as phases already support setting the Nix archive to a specific version, https://github.com/railwayapp/nixpacks/blob/7653b2f1a8b7c4532adb51d1c8f7f20cbd26386c/src/nixpacks/plan/phase.rs/#L207.

See [this](https://github.com/railwayapp/nixpacks/blob/7653b2f1a8b7c4532adb51d1c8f7f20cbd26386c/src/providers/go.rs/#L65) as an example

src/nixpacks/nix/mod.rs
```diff
@@ -13,6 +13,9 @@ pub const NIXPKGS_ARCHIVE: &str = "5148520bfab61f99fd25fb9ff7bfbb50dad3c9db";
 // Version of the Nix archive that uses OpenSSL 1.1
 pub const NIXPACKS_ARCHIVE_LEGACY_OPENSSL: &str = "a0b7e70db7a55088d3de0cc370a59f9fbcc906c3";
 
+// Version of the Nix archive with the latest Deno
+pub const NIXPACKS_ARCHIVE_LATEST_DENO: &str = "734af41a2b6a21fb9bf70d9f170563b6932364bb";
```
coffee-cup, line 17: This should be moved to the deno provider file

PR #1245, opened 2024-12-20
Add support for bun.lock

djobbo:
This PR adds support for Bun's new text lockfile `bun.lock`.

Bun v1.39 added a new text based lockfile format, without this change, apps using this new format will be detected as using npm instead of bun.

PR: https://github.com/oven-sh/bun/issues/11863
More info: https://bun.sh/blog/bun-lock-text-lockfile

### DIFF
--- a/src/providers/node/mod.rs
+++ b/src/providers/node/mod.rs
@@ -362,7 +362,7 @@ impl NodeProvider {
             pkg_manager = "pnpm";
         } else if app.includes_file("yarn.lock") {
             pkg_manager = "yarn";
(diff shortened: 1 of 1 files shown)


djobbo:
I'm commenting here so the PR is not marked as stale.
This is an important issue since this lockfile format will become new the default in upcoming Bun versions.

brody192:
Thank you for the PR!

Railway is currently on a Christmas / New Years break, we will review this PR once we are back at work!

coffee-cup (approved):

PR #1249, opened 2024-12-29
fix(plan): stdout logic for valid output files

### LINKED ISSUE #1241: `nixpack plan` generates an invalid toml file for python projects containing a `.tool-versions` file.

### Describe the bug

When running `nixpack plan -f toml` in a directory containing a Poetry project and a .tool-versions file nixpacks outputs a line before it outputs the .toml-file:

```
Using poetry version from .tool-versions: 1.8.5
```

This breaks the deployment code in Coolify, because Coolify will read the output and try to parse it as toml.
I expect the same problem to occur with the json output format.

### To reproduce

1. Create a Poetry project or simulate this with `touch pyproject.toml poetry.lock`
2. `echo "poetry 1.8.5" > .tool-versions`
3. `nixpacks plan -f toml  .`

### Expected behavior

I would expect nixpacks to output just the toml file of the build plan.

### Environment

On the host:
```
$ docker -v
Docker version 27.0.3, build 7d4bcd8
```

In the container:
```
cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.20.3
PRETTY_NAME="Alpine Linux v3.20"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

``` 
$ nixpacks --version
nixpacks 1.29.0
```

CoderJoshDK:
Don't let build plans generate anything invalid files. If an error or warning should be generated, it should be outputted to stderr. That way, consumers can be pretty confident that the output is only ever in valid TOML/JSON format.

This closes #1241 and is a slightly more complete solution over #1242.

Considerations were made to have a "verbose mode" however, doing that would require significant refactoring, all for only 3 or 4 impacted providers. Instead, recoverable errors should be directed to stderr using `eprintln!`


coffee-cup (approved):
Nice, I like this solution

PR #1251, opened 2025-01-02
node env clarification

iloveitaly:
quick doc improvement, this one confused me.


coffee-cup (approved):

Suall1969:
I found this to be confusing as well, however isn't it the other way around?
From the CLI reference:
https://github.com/railwayapp/nixpacks/blob/2f5bbf818cf28db2b47697d6cdea8782f0f1b6ea/docs/pages/docs/cli.md?plain=1#L33
Runtime env vars (when the container is running) is set by the underlying container runtime e.g. `docker run -e`.

PR #1252, opened 2025-01-05
fix phoenix/elixir recipe

nbw:
# Summary

- :one: Lock file problem when deploying a Phoenix Elixir app
- :two: Improvements for the Phoenix/Elixir deploying recipe
- :three: Fixes issue with copied files for the start phase
-  :three: Fixes issue with default values for ARGS in the docker file

:wave: **the FINAL dockerfile output is in the reference section** ⬇️ 

## :one: Lock file problem

When deploying a Phoenix Elixir app to Railway there is a conflict with the lock file that cause the error: ` lock outdated: the lock is outdated compared to the options in your mix.exs`[1]. This is because the we're copying too many files during multiple phases. The solution is to update the elixir dockerfile builder to only copy specific files. 

https://elixirforum.com/t/docker-build-started-failing-unchecked-dependencies-for-environment-prod/28844/6

The [Railway template for phoenix](https://github.com/railwayapp-templates/phoenix) got around this problem by removing the lock file, which is not behaviour we want. 

## :two: General updates to the Elixir build

- run `mix ecto.setup` if ecto_sql is detected rather than postgrex
   - postgrex is a postgres specific adapter, but there are other adapters (for mysql, etc). detecting `ecto_sql` is more generic
- only copy the necessary folders and assets
- build hex, rebar, and other deps using `MIX_ENV=prod`

### Future considerations

The way we run elixir is with `mix phx.server` but actually creating a `mix release` is the better approach (it creates a binary that can be run). [More details here about why release are better](https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-why-releases). In the future, we should move away from `mix phx.server` and use releases instead, but it's beyond the scope of this PR. [Here's an example of an Elixir dockerfile that uses releases](https://github.com/superfly/flyctl/blob/2e7e50db83543036a920499d6dba43b93fb4f085/scanner/templates/phoenix/Dockerfile#L96).

## :three: Other fixes

- Issue: setting `only_include_files` in the start phase was ignored. Instead `COPY . /app` was always added to the generated dockerfile
   - ✅  if `only_include_files` is set in the start phase, use it
- Issue: default environment values aren't set in the Dockerfile
  - ✅ set default values for docker ARGs. Example output: `ARG ELIXIR_ERL_OPTIONS="+fnu" LANG="en_US.UTF-8" LANGUAGE="en_US:en" MIX_ENV="prod" NIXPACKS_METADATA="elixir"`  

### [1] Lock file problem

```
[+] Building 113.3s (15/18)                                                 docker:orbstack
 => [internal] load build definition from Dockerfile                                   0.0s
 => => transferring dockerfile: 858B                                                   0.0s
 => [internal] load metadata for ghcr.io/railwayapp/nixpacks:ubuntu-1733184274         0.9s
 => [internal] load .dockerignore                                                      0.0s
 => => transferring context: 2B                                                        0.0s
 => [ 1/14] FROM ghcr.io/railwayapp/nixpacks:ubuntu-1733184274@sha256:91b1c0ad82b8a7a  0.0s
 => [internal] load build context                                                      0.6s
 => => transferring context: 82.02MB                                                   0.5s
 => CACHED [ 2/14] WORKDIR /app/                                                       0.0s
 => CACHED [ 3/14] COPY .nixpacks/nixpkgs-c5702bd28cbde41a191a9c2a00501f18941efbd0.ni  0.0s
 => [ 4/14] RUN nix-env -if .nixpacks/nixpkgs-c5702bd28cbde41a191a9c2a00501f18941ef  102.7s
 => [ 5/14] COPY . /app/.                                                              0.9s
 => [ 6/14] RUN  mix local.hex --force                                                 2.0s
 => [ 7/14] RUN  echo test                                                             0.2s
 => [ 8/14] RUN  mix local.rebar --force                                               1.7s
 => [ 9/14] RUN  mix deps.get --only prod                                              2.5s
 => [10/14] COPY . /app/.                                                              1.1s
 => ERROR [11/14] RUN  mix compile                                                     0.4s
------
 > [11/14] RUN  mix compile:
0.382 Unchecked dependencies for environment prod:
0.382 * heroicons (https://github.com/tailwindlabs/heroicons.git - v2.1.1)
0.382   lock outdated: the lock is outdated compared to the options in your mix.exs. To fetch locked version run "mix deps.get"
0.384 ** (Mix) Can't continue due to errors on dependencies
------
Dockerfile:26
--------------------
  24 |     # build phase
  25 |     COPY . /app/.
  26 | >>> RUN  mix compile
  27 |     RUN  mix assets.deploy
  28 |     RUN  mix ecto.setup
--------------------
ERROR: failed to solve: process "/bin/bash -ol pipefail -c mix compile" did not complete successfully: exit code: 1

View build details: docker-desktop://dashboard/build/orbstack/orbstack/uxra2k6hpddi2mzjzo7r97fjr
Error: Docker build failed
```

### [2] Final Dockerfile

output of `nixpacks build --dockerfile .` on a phoenix elixir project that uses a database:

```
FROM ghcr.io/railwayapp/nixpacks:ubuntu-1733184274

ENTRYPOINT ["/bin/bash", "-l", "-c"]
WORKDIR /app/


COPY .nixpacks/nixpkgs-c5702bd28cbde41a191a9c2a00501f18941efbd0.nix .nixpacks/nixpkgs-c5702bd28cbde41a191a9c2a00501f18941efbd0.nix
RUN nix-env -if .nixpacks/nixpkgs-c5702bd28cbde41a191a9c2a00501f18941efbd0.nix && nix-collect-garbage -d


ARG ELIXIR_ERL_OPTIONS="+fnu" LANG="en_US.UTF-8" LANGUAGE="en_US:en" MIX_ENV="prod" NIXPACKS_METADATA="elixir"
ENV ELIXIR_ERL_OPTIONS=$ELIXIR_ERL_OPTIONS LANG=$LANG LANGUAGE=$LANGUAGE MIX_ENV=$MIX_ENV NIXPACKS_METADATA=$NIXPACKS_METADATA

# setup phase
# noop

# install phase
COPY config/config.exs /app/config/config.exs
COPY config/${MIX_ENV}.exs /app/config/${MIX_ENV}.exs
COPY mix.exs /app/mix.exs
COPY mix.lock /app/mix.lock
RUN  mix local.hex --force
RUN  mix local.rebar --force
RUN  mix deps.get --only prod

# build phase
COPY config/runtime.exs /app/config/runtime.exs
COPY assets /app/assets
COPY priv /app/priv
COPY config /app/config
COPY lib /app/lib
RUN  mix compile
RUN  mix assets.deploy





# start


CMD ["mix ecto.setup && mix phx.server"]
```

### DIFF
--- a/src/nixpacks/builder/docker/dockerfile_generation.rs
+++ b/src/nixpacks/builder/docker/dockerfile_generation.rs
@@ -148,7 +148,7 @@ impl DockerfileGenerator for BuildPlan {
                 // Pull the variables in from docker `--build-arg`
                 variables
                     .iter()
-                    .map(|var| var.0.to_string())
+                    .map(|var| format!("{}=\"{}\"", var.0.trim(), var.1.trim()))
                     .collect::<Vec<_>>()
                     .join(" "),
                 // Make the variables available at runtime
--- a/src/nixpacks/builder/docker/dockerfile_generation.rs
+++ b/src/nixpacks/builder/docker/dockerfile_generation.rs
@@ -334,16 +334,15 @@ impl DockerfileGenerator for StartPhase {
             None => String::new(),
         };
 
-        let dockerfile: String = match &self.run_image {
-            Some(run_image) => {
-                let copy_cmds = utils::get_copy_from_commands(
-                    "0",
-                    &self.only_include_files.clone().unwrap_or_default(),
-                    APP_DIR,
-                );
+        let dockerfile: String = if let Some(run_image) = &self.run_image {
+            let copy_cmds = utils::get_copy_from_commands(
+                "0",
+                &self.only_include_files.clone().unwrap_or_default(),
+                APP_DIR,
+            );
 
-                // RUN true to prevent a Docker bug https://github.com/moby/moby/issues/37965#issuecomment-426853382
-                formatdoc! {"
+            // RUN true to prevent a Docker bug https://github.com/moby/moby/issues/37965#issuecomment-426853382
+            formatdoc! {"
                   # start
                   FROM {run_image}
                   ENTRYPOINT [\"/bin/bash\", \"-l\", \"-c\"]
@@ -354,29 +353,27 @@ impl DockerfileGenerator for StartPhase {
                   {user_str}
                   {start_cmd}
                 ",
-                run_image=run_image,
-                APP_DIR=APP_DIR,
-                copy_cmds=copy_cmds.join("\n"),
-                user_str=user_str,
-                start_cmd=start_cmd,}
-            }
-            None => {
-                // Copy over app files if provided, otherwise copy all files
-                let copy_cmds = match &self.only_include_files {
-                    Some(files) => utils::get_copy_commands(&files.clone(), APP_DIR).join("\n"),
-                    None => "COPY . /app".to_string(),
-                };
+            run_image=run_image,
+            APP_DIR=APP_DIR,
+            copy_cmds=copy_cmds.join("\n"),
+            user_str=user_str,
+            start_cmd=start_cmd,}
+        } else {
+            // Copy over app files if provided, otherwise copy all files
+            let copy_cmds = match &self.only_include_files {
+                Some(files) => utils::get_copy_commands(&files.clone(), APP_DIR).join("\n"),
+                None => "COPY . /app".to_string(),
+            };
 
-                formatdoc! {"
+            formatdoc! {"
                   # start
                   {copy_cmds}
                   {user_str}
                   {start_cmd}
                 ",
-                copy_cmds=copy_cmds,
-                start_cmd=start_cmd,
-                user_str=user_str}
-            }
+            copy_cmds=copy_cmds,
+            start_cmd=start_cmd,
+            user_str=user_str}
         };
 
         Ok(dockerfile)
--- a/src/nixpacks/builder/docker/dockerfile_generation.rs
+++ b/src/nixpacks/builder/docker/dockerfile_generation.rs
@@ -361,12 +361,19 @@ impl DockerfileGenerator for StartPhase {
                 start_cmd=start_cmd,}
             }
             None => {
+                // Copy over app files if provided, otherwise copy all files
+                let copy_cmds = match &self.only_include_files {
+                    Some(files) => utils::get_copy_commands(&files.clone(), APP_DIR).join("\n"),
+                    None => "COPY . /app".to_string(),
+                };
+
                 formatdoc! {"
                   # start
-                  COPY . /app
+                  {copy_cmds}
                   {user_str}
                   {start_cmd}
                 ",
+                copy_cmds=copy_cmds,
                 start_cmd=start_cmd,
                 user_str=user_str}
             }
--- a/src/providers/elixir.rs
+++ b/src/providers/elixir.rs
@@ -45,7 +45,7 @@ impl Provider for ElixirProvider {
             build_phase.add_cmd("mix assets.deploy".to_string());
         }
 
-        if mix_exs_content.contains("postgrex") && mix_exs_content.contains("ecto") {
+        if mix_exs_content.contains("ecto_sql") && mix_exs_content.contains("ecto") {
             build_phase.add_cmd("mix ecto.setup");
         }
 
--- a/src/providers/elixir.rs
+++ b/src/providers/elixir.rs
@@ -32,27 +32,47 @@ impl Provider for ElixirProvider {
         plan.add_phase(setup);
 
         // Install Phase
-        let mut install_phase = Phase::install(Some("mix local.hex --force".to_string()));
-        install_phase.add_cmd("mix local.rebar --force");
+        let mut install_phase =
+            Phase::install(Some("MIX_ENV=prod mix local.hex --force".to_string()));
+        install_phase.only_include_files = Some(vec![
+            "config/config.exs".to_string(),
+            "config/prod.exs".to_string(),
+            "mix.exs".to_string(),
+            "mix.lock".to_string(),
+        ]);
+        install_phase.add_cmd("MIX_ENV=prod mix local.rebar --force");
         install_phase.add_cmd("mix deps.get --only prod");
         plan.add_phase(install_phase);
 
         // Build Phase
         let mut build_phase = Phase::build(Some("mix compile".to_string()));
+        build_phase.only_include_files = Some(vec![
+            "config/runtime.exs".to_string(),
+            "assets".to_string(),
+            "priv".to_string(),
+            "config".to_string(),
+            "lib".to_string(),
+        ]);
         let mix_exs_content = app.read_file("mix.exs")?;
-
         if mix_exs_content.contains("assets.deploy") {
             build_phase.add_cmd("mix assets.deploy".to_string());
-        }
-
-        if mix_exs_content.contains("ecto_sql") && mix_exs_content.contains("ecto") {
-            build_phase.add_cmd("mix ecto.setup");
         }
 
         plan.add_phase(build_phase);
 
         // Start Phase
-        let start_phase = StartPhase::new("mix phx.server".to_string());
+
+        // If the app uses Ecto, we need to run the setup command before starting the server
+        let start_phase_cmd =
+            if mix_exs_content.contains("ecto_sql") && mix_exs_content.contains("ecto") {
+                "mix ecto.setup && mix phx.server"
+            } else {
+                "mix phx.server"
+            };
+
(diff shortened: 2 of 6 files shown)


src/nixpacks/builder/docker/dockerfile_generation.rs
```diff
@@ -361,12 +361,19 @@ impl DockerfileGenerator for StartPhase {
                 start_cmd=start_cmd,}
             }
             None => {
+                // Copy over app files if provided, otherwise copy all files
+                let copy_cmds = match &self.only_include_files {
```
nbw, line 365: START PHASE: if `only_include_files` is specified, then use it. Otherwise default to `COPY . /app`

src/providers/elixir.rs
```diff
@@ -32,27 +32,47 @@ impl Provider for ElixirProvider {
...
+        let start_phase_cmd =
+            if mix_exs_content.contains("ecto_sql") && mix_exs_content.contains("ecto") {
+                "mix ecto.setup && mix phx.server"
+            } else {
+                "mix phx.server"
+            };
+
+        let mut start_phase = StartPhase::new(start_phase_cmd.to_string());
+        // Skip copying any additional files for the start phase
+        start_phase.only_include_files = Some(vec![]);
```
nbw, line 75: we don't want to copy any files before starting.

src/providers/elixir.rs
```diff
@@ -32,27 +32,47 @@ impl Provider for ElixirProvider {
...
+            "config".to_string(),
+            "lib".to_string(),
+        ]);
         let mix_exs_content = app.read_file("mix.exs")?;
-
         if mix_exs_content.contains("assets.deploy") {
             build_phase.add_cmd("mix assets.deploy".to_string());
         }
 
-        if mix_exs_content.contains("postgrex") && mix_exs_content.contains("ecto") {
```
nbw, line 48: this won't work because the DATABASE_URL isn't available at this time. [Railway's own docs](https://docs.railway.com/guides/phoenix#prepare-our-phoenix-app-for-deployment) suggesting adding `mix ecto.setup` to the start phase instead.

src/providers/elixir.rs
```diff
@@ -76,6 +96,8 @@ impl ElixirProvider {
     fn default_elixir_environment_variables() -> EnvironmentVariables {
         let var_map = vec![
             ("MIX_ENV", "prod"),
+            ("LANG", "en_US.UTF-8"),
```
nbw, line 99: LANG and LANGUAGE are required for runtime and good to set as defaults

src/providers/elixir.rs
```diff
@@ -32,27 +32,47 @@ impl Provider for ElixirProvider {
         plan.add_phase(setup);
 
         // Install Phase
-        let mut install_phase = Phase::install(Some("mix local.hex --force".to_string()));
-        install_phase.add_cmd("mix local.rebar --force");
+        let mut install_phase =
+            Phase::install(Some("MIX_ENV=prod mix local.hex --force".to_string()));
```
nbw, line 36: since BUILDER secrets can't be set (only runtime ones), prefix the command with `MIX_ENV=prod`. Same to `mix local.rebar --force` below.

nbw: There is a line in this file to set default env values:

https://github.com/railwayapp/nixpacks/blob/0fcf29805382e1b23f440b6efd1e67c528827f1f/src/providers/elixir.rs#L31

And the output in the generated dockerfile is: 

```
ARG ELIXIR_ERL_OPTIONS LANG LANGUAGE MIX_ENV NIXPACKS_METADATA
ENV ELIXIR_ERL_OPTIONS=$ELIXIR_ERL_OPTIONS LANG=$LANG LANGUAGE=$LANGUAGE MIX_ENV=$MIX_ENV NIXPACKS_METADATA=$NIXPACKS_METADATA
```

but this isn't working as probably intended? 

[Docker Best Practices: Using ARG and ENV in Your Dockerfiles](https://www.docker.com/blog/docker-best-practices-using-arg-and-env-in-your-dockerfiles/)

correct me if I'm wrong, but shouldn't the ARG section look something like:

```
ARG ELIXIR_ERL_OPTIONS="..."
ARG LANG="en_US.UTF-8" 
ARG LANGUAGE="en_US:en"
ARG MIX_ENV="prod"
ARG NIXPACKS_METADATA="..."
```

reference: the args are set [in docker_generation in these lines](https://github.com/railwayapp/nixpacks/blob/c42c9e97b57d4255606e9cbeddc1f5e1a8700b31/src/nixpacks/builder/docker/dockerfile_generation.rs#L147-L153)

src/nixpacks/builder/docker/dockerfile_generation.rs
```diff
@@ -148,7 +148,7 @@ impl DockerfileGenerator for BuildPlan {
                 // Pull the variables in from docker `--build-arg`
                 variables
                     .iter()
-                    .map(|var| var.0.to_string())
+                    .map(|var| format!("{}=\"{}\"", var.0.trim(), var.1.trim()))
```
nbw, line 151: This line changes the Dockerfile output from:

```Dockerfile
ARG ELIXIR_ERL_OPTIONS LANG LANGUAGE MIX_ENV NIXPACKS_METADATA
```

To:

```Dockerfile
ARG ELIXIR_ERL_OPTIONS="+fnu" LANG="en_US.UTF-8" LANGUAGE="en_US:en" MIX_ENV="prod" NIXPACKS_METADATA="elixir"
```

Looking at the [docker arg docs](https://docs.docker.com/reference/dockerfile/#arg) and from my own testing, this simply sets a default value for the ARG but it can be overridden by setting a variable.

nbw: without this line, the rest of the dockerfile doesn't actually inherit values like MIX_ENV for the build phases and then we need to find a different workaround (like prefixing commands with `MIX_ENV=prod mix local.rebar --force`)

src/nixpacks/builder/docker/dockerfile_generation.rs
```diff
@@ -354,22 +353,27 @@ impl DockerfileGenerator for StartPhase {
...
-            None => {
-                formatdoc! {"
+            run_image=run_image,
+            APP_DIR=APP_DIR,
+            copy_cmds=copy_cmds.join("\n"),
+            user_str=user_str,
+            start_cmd=start_cmd,}
+        } else {
+            // Copy over app files if provided, otherwise copy all files
+            let copy_cmds = match &self.only_include_files {
```
nbw, line 363: START PHASE: if only_include_files is specified, then use it. Otherwise default to COPY . /app

src/providers/gleam/mod.rs
```diff
@@ -96,8 +96,6 @@ impl GleamProvider {
     }
 
     fn get_start(&self, _app: &App, _env: &Environment) -> StartPhase {
-        let mut phase = StartPhase::new("./build/erlang-shipment/entrypoint.sh run");
-        phase.only_include_files = Some(vec!["build/erlang-shipment".into()]);
-        phase
+        StartPhase::new("./build/erlang-shipment/entrypoint.sh run")
```
nbw, line 99: [failed build](https://github.com/railwayapp/nixpacks/actions/runs/12618956113/job/35163065425?pr=1252)

the `phase.only_include_files = Some(vec!["build/erlang-shipment".into()]);` was never working to begin with (since it was fixed in this PR). We can remove the line to keep the previous behaviour.

CleanShot 2025-01-05 at 20 55 45@2x

src/nixpacks/builder/docker/dockerfile_generation.rs
coffee-cup: I believe we have intentionally not done this because then all variables will be logged in the build logs. The current way of injecting build variables prevents that.

src/nixpacks/builder/docker/dockerfile_generation.rs
coffee-cup: I believe `only_include_files` was only used when the runtime image was different from the build image. So this makes the uses the same 👍

src/nixpacks/builder/docker/dockerfile_generation.rs
nbw: but this is injecting default values of environment variables. isn't it okay for these ones to be logged? like `MIX_ENV= prod`

there shouldn't be any user specific secrets set as defaults afaik.

nbw: @coffee-cup is this the only discussion point that's blocking? I will comb through the codebase a bit and see if there are any defaults that we might handle differently. the alternative here is to add some flag to explicitly say "in this case, please fill in the defaults"

PR #1257, opened 2025-01-10
SPA provider (vite)

Milo123459:
- **vite spa provider**
- **also use out directory if specified in the build command**
- **vite spa provider**

This PR introduces a Vite SPA provider which will automatically run supported applications with Caddy.

It can be configured via NIXPACKS_SPA_CADDY to turn off Caddy, or if you want to turn Caddy on for an application that doesn't automatically get detected, you can specify a build directory with the NIXPACKS_SPA_OUT_DIR variable.

- [x]  Docs are updated if needed


Milo123459:
no idea why the lunatic one is failing

src/providers/node/mod.rs
```diff
@@ -203,9 +219,25 @@ impl Provider for NodeProvider {
         // Start
         let start = NodeProvider::get_start_cmd(app, env)?.map(StartPhase::new);
 
-        let mut plan = BuildPlan::new(&vec![setup, install, build], start);
-        plan.add_variables(NodeProvider::get_node_environment_variables());
+        let mut phases = vec![setup, install, build];
+        if let Some(caddy) = SpaProvider::caddy_phase(app, env) {
+            phases.insert(1, caddy); // insert after setup and before build
```
coffee-cup, line 224: Order in the `phases` array doesn't matter. The `caddy` phase should `depend_on` the setup and build should depend on caddy. We topologically sort the phases based on these dependencies when building.

src/providers/node/mod.rs
```diff
@@ -203,9 +219,25 @@ impl Provider for NodeProvider {
...
+        }
 
+        let mut plan = BuildPlan::new(&phases, start);
+        if SpaProvider::caddy_phase(app, env).is_some() {
```
coffee-cup, line 228: Suggestion: We call this a few times in the file. We should call it once and save it to a variable for reuse.

src/providers/node/mod.rs
```diff
@@ -203,9 +219,25 @@ impl Provider for NodeProvider {
...
+            plan.add_static_assets(static_asset_list! {
+                "Caddyfile" => include_str!("spa/Caddyfile")
```
coffee-cup, line 230: Can the Caddy provider return the static assets that are necessary. That way this top-level node provider doesn't need to know about the internal details of how to build an SPA.

src/providers/node/spa/vite.rs
```diff
@@ -0,0 +1,64 @@
...
+            || pkg.has_dependency("lit")
+            || pkg.has_dependency("solid-js")
+            || pkg.has_dependency("@builder.io/qwik")
+    }
+
+    pub fn get_output_directory(app: &App) -> String {
+        let config = app
+            .read_file("vite.config.js")
+            .or(app.read_file("vite.config.ts"));
+        let r = Regex::new(r#"outDir:\s*['"`](.*?)['"`]"#).unwrap();
```
coffee-cup, line 42: Can you add some tests in this file for the regex and these checks. I'm a little concerned of accidentally considering apps as SPAs when the shouldn't be.

coffee-cup (commented):
Nice work! This will definitely make it easier for people coming from Vite apps.

Do you know how this will interact with users who have a custom nixpacks.toml and Caddy file for deploying static sites already?

Milo123459:
I'm not sure, but the Caddyfile wouldn't overwrite the users. And im pretty sure it would just use what's in nixpacks.toml as opposed to what the SPA provider generates.

src/providers/node/mod.rs
```diff
@@ -203,9 +219,24 @@ impl Provider for NodeProvider {
         // Start
         let start = NodeProvider::get_start_cmd(app, env)?.map(StartPhase::new);
 
-        let mut plan = BuildPlan::new(&vec![setup, install, build], start);
-        plan.add_variables(NodeProvider::get_node_environment_variables());
+        let mut phases = vec![setup, install, build];
+        if let Some(caddy) = SpaProvider::caddy_phase(app, env) {
+            phases.push(caddy); // insert after setup and before build
```
coffee-cup, line 224: Nit: remove the comment as it is not relevant anymore.

Does the build phase need to be updated to depend on the caddy? Or does it not matter the ordering?

src/providers/node/mod.rs
```diff
@@ -299,6 +330,13 @@ impl NodeProvider {
             }
         }
 
+        if SpaProvider::caddy_phase(app, env).is_some() {
+            return Ok(Some(format!(
+                "exec caddy run --config {} --adapter caddyfile 2>&1",
```
coffee-cup, line 335: The start command should live inside the SpaProvider to keep things isolated

src/providers/node/spa/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+            if s.to_lowercase() == "false" || s.to_lowercase() == "0" {
+                return None;
+            }
+        }
+        if Self::is_spa(app)
+            && (vite::ViteSpaProvider::caddy_allowlist(app)
+                || env.get_config_variable("SPA_OUT_DIR").is_some())
+        {
+            let mut caddy = Phase::new("caddy");
+            caddy.set_nix_archive(String::from("ba913eda2df8eb72147259189d55932012df6301")); // caddy 2.0.4
```
coffee-cup, line 29: Nit: extract this to a constant at the top

src/providers/node/spa/vite.rs
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                }
+            }
+        }
+        String::from("dist")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
coffee-cup, line 67: Can you also add some tests for examples that shouldn't match. Like NextJS or Express

src/providers/node/mod.rs
Milo123459: it doesn't matter

coffee-cup (approved):
Looks good! I've tested on various repos and it works well 🎉

PR #1271, opened 2025-01-21
feat: more py deps for audio and image processing

iloveitaly:
- Integrate `ffmpeg-headless` for handling audio dependencies when using `pydub` or `pymovie`.
- Include `poppler_utils` to support PDF processing for `pdf2image`.


coffee-cup (approved):
🎉

coffee-cup:
Could you please also add a test that uses these dependencies. I'm also not sure we should be adding something custom for pymovie. As far as I can tell it is a very small package with little documentation?

iloveitaly:
> Could you please also add a test that uses these dependencies. I'm also not sure we should be adding something custom for pymovie. As far as I can tell it is a very small package with little documentation?

yeah we are using pydub, pymovie was one we looked into briefly so figured i'd add it but didn't realize it was so small. Will remove.

iloveitaly:
@coffee-cup added some tests

coffee-cup (approved):

Issue #1064, opened 2024-02-07
Output environment variables during build


iloveitaly:
### Feature request

Adding a section to this display:

<img width="722" alt="image" src="https://github.com/railwayapp/nixpacks/assets/150855/7d6b5124-9e7f-43f7-9cc5-757bfd4030f9">

### Motivation

Inform developers about the default set of environment variables in the build. This would make it easy for them to understand which variables might be worth overriding to a different value, as well as easily introspecting what dynamic defaults were chosen for particular variables.

Right now the only way to investigate this is to Docker inspect the resulting image and look through the massive JSON output.

Related to: https://github.com/railwayapp/nixpacks/pull/1063

### Contribution

I can add this.

PR #1273, opened 2025-01-24
Print environment variables

sinclairnick:
This PR adds environment variables to the build output a la https://github.com/railwayapp/nixpacks/issues/1064.

I haven't added tests as I'm not sure if they're needed, but I can do if need be.


coffee-cup (commented):
What does the output look like?

This should also be optional since variables can be secret and we are purposefully not outputting them in the build logs.

What is the use case for this?

coffee-cup:
Going to close this PR because I don't think we should be printing variables and secrets to stdout

Issue #1302, opened 2025-04-10
Building react-router v7 applications


fattila16:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The `node` prodiver of nixpacks is quite complex, as it has to support a lot of different use cases.

There is a dedicated `spa.rs` file where the logic of figuring out whether the project is an SPA is discovered.

In the `1.34.1` release there was a fix for building `remix` applications.

Since 2024-11-21 remix has been merged with `react-router` as v7 was released.

The fix that was introduced for remix, only works for remix. Remix is now maintained in the react-router project.

If someone would like to build a react-router v7 project, nixpacks will think it's an SPA, nevertheless that it might not be.

A quick fix can be that in the spa mode one includes another check in the conditions to look for `@react-router/node`.

However that won't necessarily handle when the user is in `framework` mode but directly disables the SSR in the `react-router.config.ts` file.

Ultimately I think there should be a more robust check, as `react-router` can be used in multiple modes:
- framework
- data
- declarative

Reference: https://reactrouter.com/start/modes

### To reproduce

Steps to reproduce the bug:

1. Setup a react-router v7 project in framework mode
   ```shell
   npx create-react-router@latest my-react-router-app
   ```
2. Run `nixpacks plan .`
3. It will start the application with caddy

### Expected behavior

It should not handle react-router v7 apps as an SPA, unless the react-router app is configured to be an SPA either in framework mode or library mode.

### Environment

Operating System: macOS 15.3.2
Nixpacks version: 1.35.0

PR #1303, opened 2025-04-10
Fix/fix react router v7 application building

fattila16:
This PR is fixing the issue described in: https://github.com/railwayapp/nixpacks/issues/1302


fattila16:
A clippy error was fixed in this PR too. A more important issue is that the `rust-ring` example was failing. I don't have that much rust knowledge, I check out the rust provider and saw that in case of `musl` is needed two nix pkgs are installed. Namely: `musl` and `musl.dev`. The later, `musl.dev` I can't seem to find it on the nixos package registry (Maybe I wasn't searching in the correct place?). Therefor I've added the `musl-tools` apt package. This seems to fix this issue.

Issue #1301, opened 2025-04-08
[FEAT] Allow user to specify a preferred package manager


lmmx:
Hi there, I'm facing a bit of a blocker here due to a bug due to the poetry installation story and a v2 deprecation (it happens) but can see a way out that I would otherwise be able to use but can't due to how the Python provider setup has been coded:

https://github.com/railwayapp/nixpacks/blob/e8a1326a3e47834a54fd44dbaa51a9833c04f071/src/providers/python.rs#L208-L271

As a feature request (which I'd be happy to contribute, it should be easy), I'd like to be able to specify a particular package manager rather than have this hardcoded order of preference. That is: I would like to say "use uv not poetry, even if I have a poetry.lock lockfile". In my case I am deploying on another provider with poetry and so can't delete that poetry lockfile yet, but I'd like to (and in fact have to) use uv to deploy to Railway.

So I think what should be permitted is to short circuit this check with a hardcoded preference, falling back to the conditionals if not set.

iloveitaly:
@lmmx We are deprecating nixpacks and replacing it with [railpack](https://github.com/railwayapp/railpack).

With railpack, you [can define exact package manager versions](https://github.com/iloveitaly/python-starter-template/blob/711aa64d5782e12169094ec0e25488fa9d8eed49/.tool-versions#L6) with mise and/or .tool-versions.

Could you give railpack a shot and open up an issue there if you run into any issues?

PR #1304, opened 2025-04-10
refactor(providers): specify `NIXPACKS_PYTHON_PACKAGE_MANAGER`

lmmx:
This PR refactors the Python rewrite package manager choice logic using an enum to make the decision tree explicitly clear, and thus also to allow a particular branch to be specified in config

- Motivated by a wish (#1301) to allow for the case where a Python project may have multiple valid lock files, and to override the default preference, e.g. to reverse the default of `Poetry.lock` > `uv.lock`

## Tasks

- refactors the Python providers module taking care to preserve the exact pre-existing behaviour (so no existing projects will choose a different package manager to install from)
- [x] Tests are added/updated
  - Unit tests to correctly resolve config variable for `NIXPACKS_PYTHON_PACKAGE_MANAGER` as "poetry", "uv", etc.
- [x] Docs are updated

### DIFF
--- a/docs/pages/docs/providers/python.md
+++ b/docs/pages/docs/providers/python.md
@@ -38,41 +38,59 @@ You also specify the exact poetry, pdm, and uv versions:
 - The `NIXPACKS_PDM_VERSION` environment variable
 - The `NIXPACKS_UV_VERSION` environment variable or `uv` in a `.tool-versions` file
 
+You can specify a particular package manager, to override the lockfile-based choice, by setting the
+`NIXPACKS_PYTHON_PACKAGE_MANAGER` environment variable to one of the following:
+
+- `auto` to choose based on the available lockfiles (default)
+- `requirements` to install using `pip` from `requirements.txt`
+- `setuptools` to install using `pip` with `build` and `setuptools`
+- `poetry` to install using `poetry` from `poetry.lock`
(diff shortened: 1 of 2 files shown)


madisvain:
Currently, when both requirements.txt and uv.lock exist in a project, the system automatically prioritizes pip (via requirements.txt) because of the fixed priority order in the code. I need to keep requirements.txt in my project for general pip compatibility, but specifically on Railway I'd like to try using uv instead. This PR would allow overriding the default package manager selection through the NIXPACKS_PYTHON_PACKAGE_MANAGER configuration variable, making it possible to choose uv even when requirements.txt is present.

coffee-cup (approved):
This looks great! ty

lmmx:
yw, will take a look at the clippy lint failure my bad sorry!

lmmx:
Reviewable once more :saluting_face:

coffee-cup:
Just a few more linting failures. You should be able to fix these automatically (or see where they are) with `cargo lint-fix` locally. Make sure you are using the latest Rust version.

lmmx:
I just see one, think I got it!

lmmx:
> This pull request is stale because it has been open 10 days with no activity. Remove stale label or comment or this will be closed in 5 days

Ready to review, not stale! @coffee-cup

PR #1305, opened 2025-04-17
bun: 1.1.20 -> 1.2.10

Xiloe:
This PR updates bun from 1.1.20 which is really old to 1.2.10 which is the latest version that released yesterday.


coffee-cup:
Closing in favour of https://github.com/railwayapp/nixpacks/pull/1312

PR #1347, opened 2025-07-24
feat: add collectstatic to Django start

jhthompson:
This PR adds the `collectstatic` Django management command to the `start` phase for Django apps.

This is extremely common for Django apps. See:
- https://docs.djangoproject.com/en/5.2/ref/contrib/staticfiles/#collectstatic 
- https://www.loopwerk.io/articles/2025/coolify-django/


jhthompson:
This might not actually be the best idea. Not sure if 100% of Django apps need this.

PR #1352, opened 2025-08-09
Update Gleam provider to install version chosen in main toml

h22roscoe:
This PR helps avoid the guesswork of basing the gleam version off the stdlib version which doesn't really work. Also updates the Nixpks archive version to install a more modern version of erlang dep which the current gleam stdlib requires


h22roscoe:
Can anyone help with this? It's working for me and I've been using this already but needs a repo owner to add a label

PR #1355, opened 2025-08-12
Adds nx 20 support

ntgussoni:
Since nx 20, project.json is optional, and instead project configurations can be placed in the package.json under the nx property.

This makes NX detection fail on these cases, which are the de-facto setup for newer NX versions.
See: https://github.com/nrwl/nx/issues/30464#issuecomment-2747498463

This PR addresses that and improves NX detection, while maintaining the legacy way of doing it.

Please note that im not a Rust guy, so all feedback is more than welcome.

It's #1322 but reopened


ntgussoni:
Hi @coffee-cup I'm back after months with this fix after our discussion in #1322 . I hope this is sufficient now and i'll gladly test it once it's released

ntgussoni:
@coffee-cup sorry to bother you again, is it possible to get this checked? 🙏

src/providers/node/nx.rs
```diff
@@ -61,20 +71,81 @@ impl Nx {
...
-            return nx_json.default_project;
+            if let Some(default_project) = nx_json.default_project {
+                return Some(default_project);
+            }
+        }
+
+        // Third, try to auto-detect by looking for apps with valid configurations
+        if app.includes_directory("apps") {
+            // Look for directories in apps/ that have either project.json or package.json with nx config
+            if let Ok(entries) = std::fs::read_dir(app.source.join("apps")) {
```
coffee-cup: We should use the methods on the app instead of reading with fs directly here. You can use `find_glob` ([ref](https://github.com/railwayapp/nixpacks/blob/main/src/nixpacks/app.rs#L73))

coffee-cup (commented):
Sorry about the delayed review. Logic is looking good, but can you use the methods on app instead of calling fs directly

ntgussoni:
> Sorry about the delayed review. Logic is looking good, but can you use the methods on app instead of calling fs directly

Done! Thanks for your time

coffee-cup (approved):

PR #1378, opened 2025-11-12
docs: add maintenance mode notice and recommend Railpack

iloveitaly:
Updated all high-traffic documentation pages to include a prominent maintenance mode warning that recommends Railpack as the replacement for Nixpacks. This ensures users are aware of the project's status and are directed to the actively maintained alternative.


iloveitaly:
@coffee-cup the nixpacks docs don't indicate it's in maint mode.

coffee-cup (approved):

PR #1426, opened 2026-06-06
fix: remove unused clap arg import in main.rs

enlorik:
Removes stale arg import left in the clap use statement — clippy unreadable/unused import warning.


enlorik:
Withdrawing this — on reflection it's a trivial change that isn't worth maintainer review time. Thanks for the project!

